PR #4, opened 2019-08-08
Update to labeler v2

shawnnapora:
This updates labeler to use the actions v2 action type


blesnatalman:
> This updates labeler to use the actions v2 action type

PR #5, opened 2019-08-14
Fix typo in README.md

flexzuu:
Based on this resource: https://help.github.com/en/articles/events-that-trigger-workflows#webhook-events


damccorm:
Good catch, thanks for the contribution!

PR #8, opened 2019-08-16
Fetch configuration content from PR ref

### LINKED ISSUE #6: HttpError: Not Found when run with new yml config
I tried to use the labeler with the new YML syntax and it seems there is an issue on how the labeler is published. I always get the following msg.
I tested the `v2` and `v2.0.0` releases

```
Run actions/labeler@v2.0.0
  with:
    repo-token: ***
    configuration-path: .github/labeler.yml
##[error]HttpError: Not Found
##[error]Not Found
##[error]Node run failed with exit code 1
```

```.github/workflows/labeler.yml
name: "Pull Request Labeler"
on: pull_request

jobs:
  label:
    runs-on: ubuntu-latest
    steps:
    - uses: actions/labeler@v2.0.0
      with:
        repo-token: "${{ secrets.GITHUB_TOKEN }}"
```

eebs:
Fixes #6

When fetching the configuration content from the file specified by `configuration-path`, we should load the content from the ref of the PR.

By default, `repos.getContents` will [use the repository's default branch][1] if no `ref` is specified.

When first configuring this action, assuming the configuration file is added in the same PR as the action, the configuration file will not exist on the default branch. This causes an `HttpError: Not Found` to occur when attempting to load the configuration from the default branch.

This commit updates the call to `getContents` to add the `ref` parameter, set to the `GITHUB_SHA` value from the context. This will cause the action to load the contents of the configuration file from the ref of the PR, rather than the default branch.

---

This PR also updates the packages to use version 1.0.0 of all the toolkit action packages, as they were using local files before.

I noticed that `src/main.ts` is not formatted according to the prettier rules. I'm happy to open another PR that applies prettier to the current source. I had to disable prettier during implementation of this commit in order to keep the changes relevant to this PR.

[1]: https://developer.github.com/v3/repos/contents/#get-contents

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,11 +22,11 @@
   "author": "GitHub",
   "license": "MIT",
   "dependencies": {
-    "@actions/core": "file:toolkit/actions-core-0.0.0.tgz",
-    "@actions/exec": "file:toolkit/actions-exec-0.0.0.tgz",
-    "@actions/github": "file:toolkit/actions-github-0.0.0.tgz",
-    "@actions/io": "file:toolkit/actions-io-0.0.0.tgz",
-    "@actions/tool-cache": "file:toolkit/actions-tool-cache-0.0.0.tgz",
+    "@actions/core": "^1.0.0",
+    "@actions/exec": "^1.0.0",
+    "@actions/github": "^1.0.0",
+    "@actions/io": "^1.0.0",
+    "@actions/tool-cache": "^1.0.0",
     "js-yaml": "^3.13.1",
     "minimatch": "^3.0.4",
     "semver": "^6.1.1"
--- a/src/main.ts
+++ b/src/main.ts
@@ -92,7 +92,8 @@ async function fetchContent(
   const response = await client.repos.getContents({
     owner: github.context.repo.owner,
     repo: github.context.repo.repo,
-    path: repoPath
+    path: repoPath,
+    ref: github.context.sha
   });
 
   return Buffer.from(response.data.content, 'base64').toString();
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,36 +1,41 @@
 {
-  "name": "node12-template-action",
-  "version": "0.0.0",
+  "name": "labeler",
+  "version": "2.0.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
     "@actions/core": {
-      "version": "file:toolkit/actions-core-0.0.0.tgz",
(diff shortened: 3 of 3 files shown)


dougnukem:
Why doesn't the action access the configuration from the local workspace git `checkout` rather than going to the github API to fetch the file?

eebs:
> Why doesn't the action access the configuration from the local workspace git `checkout` rather than going to the github API to fetch the file?

There actually isn't a local checkout of the repository. This action only uses the event payload and the API to label the PR.

package.json
```diff
@@ -22,11 +22,11 @@
   "author": "GitHub",
   "license": "MIT",
   "dependencies": {
-    "@actions/core": "file:toolkit/actions-core-0.0.0.tgz",
-    "@actions/exec": "file:toolkit/actions-exec-0.0.0.tgz",
-    "@actions/github": "file:toolkit/actions-github-0.0.0.tgz",
-    "@actions/io": "file:toolkit/actions-io-0.0.0.tgz",
-    "@actions/tool-cache": "file:toolkit/actions-tool-cache-0.0.0.tgz",
+    "@actions/core": "^1.0.0",
+    "@actions/exec": "^1.0.0",
```
damccorm: Can you actually just cut exec, io, and tool-cache? We don't need them in this action,

src/main.ts
```diff
@@ -92,7 +92,8 @@ async function fetchContent(
   const response = await client.repos.getContents({
```
damccorm: Do we maybe want to retry on the default branch if this 404's? The scenario I'm concerned about is:
1) User X branches repo
2) This action gets merged into master
3) User X PRs, but is out of sync with master

No label would get applied at this point if they hadn't synced with master yet, even if they sync. I know this is kinda an edge case, but its most likely to crop up the first couple of times the action is run and isn't a great first experience.

package.json
eebs: Extra packages removed in 8bc948481e24a0a64db20c4677dc7ba203481511

src/main.ts
eebs: I'm not sure I'd agree with retrying on the default branch. My feeling is that workflows/actions should only run against the code in the ref/sha they are committed to. I think I'd be surprised if my branch used actions/configuration from a different branch.

I think the scenario you described is fairly unlikely and has an obvious remedy (rebase/merge master).

What do you think of shipping this as-is, and making the change if we get feedback from the community?

damccorm:
Published!

Issue #9, opened 2019-08-21
Support for exclusion paths


jalaziz:
There are cases where we want to match an entire directory but not certain subdirectories.

Instead of having to list each matching subdirectory, it would be nice to be able to exclude some paths from a match.

An example may be:
```
label1:
- match: "example1/**/*"
  exclude: ["example1/foo/*", "example1/bar/*"]
```

damccorm:
This seems like a helpful idea! I think we probably should be able to do something like this, I want to be careful here though - I don't want to break existing configurations, so I don't want to change the syntax of the `yml` file.

I think a better approach might be to support `!` as a negate pattern. So you're example would look the following instead:
```
label1:
- "!example1/foo/*"
- "!example1/bar/*"
- "example1/**/*"
```

Probably, we'd want to evaluate sequentially until we find a pattern that matches. So that way we could even do something like:

```
label2:
- "!example2/foo/bar"
- "example2/foo/**/*"
- "!example2/**/*"
- "**/*"
```

The above example would not match `example2/foo/bar/abc.txt`, would match `example2/foo/baz/abc.txt`, would not match `example2/notFoo/baz/abc.txt`, and would match `notExample2/foo/baz/abc.txt` since the patterns are evaluated in order - I think this would be about as powerful as we can get.

Thoughts?

jalaziz:
> I don't want to break existing configurations, so I don't want to change the syntax of the yml file.

Actually, that's exactly why I proposed the syntax above. Maps aren't currently handled and it should be possible to support current configuration files without breaking them.

I had thought about negation, but my fear is that it would cause confusion with the semantics of the list as it stands today. Today the label is applied if _any_ path matches. Should negation mean that a non-match should receive the label? Or that we should stop searching? It would also mean that the order of the rules matter. That's why I figured it would be better to be explicit with a map.

That being said, if you're comfortable with order being significant, then I'd certainly prefer your proposed syntax. It's certainly simpler than what I proposed!

damccorm:
> Actually, that's exactly why I proposed the syntax above. Maps aren't currently handled and it should be possible to support existing configuration files without breaking them.

Ah I see, that makes sense - at the same time, I don't think I want us to support 2 different syntaxes if possible though - that leads me to prefer an additive pattern if possible

> if you're comfortable with order being significant

I actually think this is a pretty common pattern that I've seen elsewhere. I think as long as we doc it, it should be fine. I think it makes intuitive sense as well (at least to me 😄).

[...]

PR #22, opened 2019-09-14
Add support for exclusions

jalaziz:
Paths can be negated to stop searching through the remaining patterns in
a the glob list. All changed files tested and at least one matching file
will result in a label being added.

Fixes actions/labeler#9

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,35 @@ Note that only pull requests being opened from the same repository can be labele
 
 Create a `.github/labeler.yml` file with a list of labels and [minimatch](https://github.com/isaacs/minimatch) globs to match to apply the label.
 
-The key is the name of the label in your repository that you want to add (eg: "merge conflict", "needs-updating") and the value is the path (glob) of the changed files (eg: `src/**/*`, `tests/*.spec.js`)
+The key is the name of the label in your repository that you want to add (eg: "merge conflict", "needs-updating") and the value is the path (glob) of the changed files (eg: `src/**/*`, `tests/*.spec.js`) or a match object.
+
+#### Match Object
+
+For more control over matching, you can provide a match object instead of a simple path glob. The match object is defined as:
+
+```yml
+- any: ['list', 'of', 'globs']
+  all: ['list', 'of', 'globs']
+```
+
+One or both fields can be provided for fine-grained matching. Unlike the top-level list, the list of path globs provided to `any` and `all` must ALL match against a path for the label to be applied.
+
+The fields are defined as follows:
+* `any`: match ALL globs against ANY changed path
+* `all`: match ALL globs against ALL changed paths
+
+A simple path glob is the equivalent to `any: ['glob']`. More specifically, the following two configurations are equivalent:
+```yml
+label1:
+- example1/*
+```
+and
+```yml
+label1:
+- any: ['example1/*']
+```
+
+From a boolean logic perspective, top-level match objects are `OR`-ed together and indvidual match rules within an object are `AND`-ed. Combined with `!` negation, you can write complex matching rules.
 
 #### Basic Examples
 
@@ -39,17 +67,15 @@ repo:
 test:
   - src/**/*.spec.js
 
-# Add 'source' label to any change to src files within the source dir EXCEPT for the build sub-folder
+# Add 'source' label to any change to src files within the source dir EXCEPT for the docs sub-folder
 source:
-- !src/build/*
-- src/**/*
+- any: ['src/**/*', '!src/docs/*']
+
+# Add 'frontend` label to any change to *.js files as long as the `main.js` hasn't changed
+frontend:
+- any: ['src/**/*.js']
+  all: ['!src/main.js']
 ```
-
-#### Exclusions
-
-Exclusions are supported by preceding the pattern with `"!"`.
-
-Patterns are evaluated in order and the first match will result in the label being assigned. If an exclusion rule matches, all remaining patterns are skipped for the path being evaluated.
 
 ### Create Workflow
 
--- a/README.md
+++ b/README.md
@@ -38,7 +38,18 @@ repo:
 # Add 'test' label to any change to *.spec.js files within the source dir
 test:
   - src/**/*.spec.js
+
+# Add 'source' label to any change to src files within the source dir EXCEPT for the build sub-folder
+source:
+- !src/build/*
+- src/**/*
 ```
+
+#### Exclusions
+
+Exclusions are supported by preceding the pattern with `"!"`.
+
+Patterns are evaluated in order and the first match will result in the label being assigned. If an exclusion rule matches, all remaining patterns are skipped for the path being evaluated.
 
 ### Create Workflow
 
--- a/dist/index.js
+++ b/dist/index.js
@@ -421,6 +421,13 @@ module.exports = new Schema({
 
 /***/ }),
 
+/***/ 34:
+/***/ (function(module) {
+
+module.exports = require("https");
+
+/***/ }),
+
 /***/ 39:
 /***/ (function(module) {
 
@@ -2278,21 +2285,13 @@ const windowsRelease = release => {
 
 	const ver = (version || [])[0];
 
-	// Server 2008, 2012, 2016, and 2019 versions are ambiguous with desktop versions and must be detected at runtime.
+	// Server 2008, 2012 and 2016 versions are ambiguous with desktop versions and must be detected at runtime.
 	// If `release` is omitted or we're on a Windows system, and the version number is an ambiguous version
 	// then use `wmic` to get the OS caption: https://msdn.microsoft.com/en-us/library/aa394531(v=vs.85).aspx
-	// If `wmic` is obsoloete (later versions of Windows 10), use PowerShell instead.
-	// If the resulting caption contains the year 2008, 2012, 2016 or 2019, it is a server version, so return a server OS name.
+	// If the resulting caption contains the year 2008, 2012 or 2016, it is a server version, so return a server OS name.
 	if ((!release || release === os.release()) && ['6.1', '6.2', '6.3', '10.0'].includes(ver)) {
-		let stdout;
-		try {
-			stdout = execa.sync('powershell', ['(Get-CimInstance -ClassName Win32_OperatingSystem).caption']).stdout || '';
-		} catch (_) {
-			stdout = execa.sync('wmic', ['os', 'get', 'Caption']).stdout || '';
-		}
-
-		const year = (stdout.match(/2008|2012|2016|2019/) || [])[0];
-
+		const stdout = execa.sync('wmic', ['os', 'get', 'Caption']).stdout || '';
+		const year = (stdout.match(/2008|2012|2016/) || [])[0];
 		if (year) {
 			return `Server ${year}`;
 		}
@@ -3489,7 +3488,7 @@ module.exports = require("child_process");
 var net = __webpack_require__(631);
 var tls = __webpack_require__(16);
 var http = __webpack_require__(605);
-var https = __webpack_require__(211);
+var https = __webpack_require__(34);
 var events = __webpack_require__(614);
 var assert = __webpack_require__(357);
 var util = __webpack_require__(669);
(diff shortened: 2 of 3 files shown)


jalaziz:
@damccorm Just a friendly ping on this PR.

bryanmacfarlane:
Can we add unit tests?

jalaziz:
I originally intended to, but there are no unit tests as far as I could tell. It didn't seem like this PR was the appropriate place to bootstrap unit tests for the project.

bryanmacfarlane:
Yeah, you're correct.  I glanced at the test folder but it has a todo ;) . https://github.com/actions/labeler/blob/master/__tests__/main.test.ts

jalaziz:
Yup, same thing I saw.

I could still give it a shot though, but fair warning, I'm fairly new to TS and have never used Jest.

TimonVS:
@jalaziz I wrote tests for a similar Action which might give you some inspiration: https://github.com/TimonVS/pr-labeler-action/pull/16/files#diff-74e406fb3eb7c98388edeb1b1c5dcf08.
I'm happy to help out wherever possible :)

ofek:
Can this be merged? 😄

bryanmacfarlane:
@dakale to have a look after the new year.

dakale:
Im not sure the logic for short circuiting exclusion matches makes sense. With this, they are only useful if you list them first, eg, if I have simply have:

```
label:
- src/*
- !/src/some_file
```

Imo, I expect a PR that changes files in src/ but also changes src/some_file, to not match this label. I would read this as "any changes to src/ but not to src/some_file"

Previously, returning `true` makes sense because you only needed at least one positive match, but with exclusions, I would think you want at least one positive match AND no exclusions

I do think we can continue short circuiting when an exclusion is met, because the logic for matching a label is at least one positive AND no exclusions, so no positive matches for globs after the exclusion would ever need to be considered. The only difference would be instead of returning true as soon as a file matches a glob, we would defer returning true until after all globs for that file have been considered and no exclusions were met which matches my above example

jalaziz:
> Im not sure the logic for short circuiting exclusion matches makes sense. With this, they are only useful if you list them first, eg, if I have simply have:
> 
> ```
> label:
> - src/*
> - !/src/some_file
> ```
> 
> Imo, I expect a PR that changes files in src/ but also changes src/some_file, to not match this label. I would read this as "any changes to src/ but not to src/some_file"
> 
> Previously, returning `true` makes sense because you only needed at least one positive match, but with exclusions, I would think you want at least one positive match AND no exclusions
> 
> I do think we can continue short circuiting when an exclusion is met, because the logic for matching a label is at least one positive AND no exclusions, so no positive matches for globs after the exclusion would ever need to be considered. The only difference would be instead of returning true as soon as a file matches a glob, we would defer returning true until after all globs for that file have been considered and no exclusions were met which matches my above example

Please see the discussion in #9. I had concerns about order being significant, but @damccorm suggested it should be fine.

dakale:
Im still not convinced that order should matter, given that no specific example was brought up in that conversation that actually requires ordering

Its not described, but I think what this follow example:

```
label2:
- "!example2/foo/bar"
- "example2/foo/**/*"
- "!example2/**/*"
- "**/*"
```

is expressing, is "None of the files in example2/**/*, except example2/foo/**/*", (which is the opposite of how I originally described the logic: any files in... except). This to me, isnt useful because its equivalent to *just listing the match with no exclusions*, because if you know what file you want to match, you can express that directly, instead of indirectly/inversely by subtracting all *non-matches* from the set of all possible files

In fact, I think the above example could be rewritten as:

```
label2:
- "**/*"
- "!example2/foo/bar"
- "example2/foo/**/*"
- "!example2/**/*"
```

without any loss of meaning (assuming the app logic is updated to not care about order), without impacting the actual logic of what is happening. You could put them in any order you want actually, it expresses the same thing, which is that there is at least one file in the PR that matches the label and isnt excluded. If you start with '**/*` and then start removing files via exclusions, you arrive at the same result (note this requires an implementation change to not short circuit at the first positive match)

Hopefully that makes sense and is actually correct. Basically what Im trying to boil it down to is a simpler system that can have its logic expressed simply: If a file matches any globs in a label, and is not excluded, the label is applied. This doesnt change any existing behavior, because "any match" is indistinguishable from "first match" when you dont have exclusions, only the implementation changes

Id be happy to merge this if its implemented as described above, or to continue this discussion

chabou:
@jalaziz do you need some help to update your PR to match @dakale specs?
I can make a new PR to address this if you don't have any free time.

jalaziz:
> @jalaziz do you need some help to update your PR to match @dakale specs?
> I can make a new PR to address this if you don't have any free time.

@chabou sorry, been a bit busy, but looking at fixing this today.

> without any loss of meaning (assuming the app logic is updated to not care about order), without impacting the actual logic of what is happening. You could put them in any order you want actually, it expresses the same thing, which is that there is at least one file in the PR that matches the label and isnt excluded. If you start with '**/*` and then start removing files via exclusions, you arrive at the same result (note this requires an implementation change to not short circuit at the first positive match)

@dakale I'm trying to recall all the context around ordering, but I believe the reason why I was suggested order would matter was because of the statement from https://github.com/actions/labeler/issues/9#issuecomment-524039895 of "Probably, we'd want to evaluate sequentially until we find a pattern that matches. So that way we could even do something like:"

As for the example from #9, I think precedence matters.

Also, your interpretation missed something:
```
label2:
- "!example2/foo/bar"
- "example2/foo/**/*"
- "!example2/**/*"
- "**/*"
```

Should translate to "match all files, except `example2/**`, but include `example2/foo/**` except `example2/foo/bar`". The reason why order matters here is that `!example2/**/*` would exclude the explicit match of `example2/foo/**/*`.

>  If you start with '**/*` and then start removing files via exclusions, you arrive at the same result (note this requires an implementation change to not short circuit at the first positive match)

Unfortunately, that's not true, because of `example2/foo/**/*` rule followed by the `!example2/**/*` rule. Applying the `!example2/**/` exclusion rule would override the `example2/foo/**/*` inclusion.

How do you feel about the map syntax I proposed? I think it would be more explicit and allow for additional use-cases such as https://github.com/actions/labeler/issues/9#issuecomment-586455011.

The idea with supporting a map would be that we can represent the boolean expression simply as: `(Rule 1 ^ Rule 1 exclusions) | (Rule 2 ^ Rule 2 exclusions) | (Rule 3 ^ Rule 3 exclusions)`. That is, the top-level list is always "ORed" and the individual list items are always "ANDed".

dakale:
Hey @jalaziz I think you bring up a great point, and the map syntax you propose does seem to seem solve it, and other usecases, like: 

Branch name: https://github.com/actions/labeler/issues/54

```
label1:
- match: "example1/foo"
  exclude: ["example1/bar"]
  branch: "featureA"
```

Change type: https://github.com/actions/labeler/issues/47

```
label1:
- match: "example1/foo"
  exclude: ["example1/bar"]
  on/event/action: "add/remove/edit/etc"
```

Having a rich object here, where the match path(s) are specified as a `match` key instead of assumed to be the default type of the value, allows us to implement a lot more functionality based on other properties, so Im all for it. Ive got some bandwidth currently to work on this action and would like to see it progress.

Have you thought through whether it might make sense to allow both the match/exclude properties to be `string | string[]` instead of match only being `string` and excludes only being `string[]`? For a particular label, the "matchers" array would still be ORed together, but it may allow more flexibility. 

Regardless, I think what you have proposed in https://github.com/actions/labeler/issues/9#issuecomment-619636954 looks promising as a starting point and should be extensible. Since a `string` exclude effectively just a length 1 array, and a `string[]` match is ultimately compatible with the interface, which is: `match` is just a boolean matches/doesnt match, regardless of if that means it matches all or matches any (AND vs OR). 

Either way Im happy to help get a simple form of these changes in. Do you want to work on a new PR or iterate on this one (to change to map syntax)?

jalaziz:
> Have you thought through whether it might make sense to allow both the match/exclude properties to be `string | string[]` instead of match only being `string` and excludes only being `string[]`? For a particular label, the "matchers" array would still be ORed together, but it may allow more flexibility.

Funny you mention that. When I was looking at #9 again, I was thinking the same thing. I would definitely add support for a list of strings. Might as well make it as robust as could be.

> Either way Im happy to help get a simple form of these changes in. Do you want to work on a new PR or iterate on this one (to change to map syntax)?

I'll iterate on this one. Will work on this today.

jalaziz:
Just an update. I was working on the new richer syntax and then got distracted. However, I'm back to looking at this and trying to get it ready for review.

It turns out that supporting https://github.com/actions/labeler/issues/9#issuecomment-586455011 is not quite the same as the exclusions we've been talking about. Specifically, the use case I originally brought up would match if at least one file does not match the excluded paths. However, the exclusion @fluzzi wants is more of a "reject if path matches".

Since I think there's value in both, I've been working on a change that supports both types of exclusion. I need to get some sleep, but hope to have it ready for review tomorrow.

jalaziz:
@dakale Still need to update the documentation, but I think what I ended up with is far more flexible than my original idea.

Each "matcher" now accepts two possible fields: `all` and `some`. Each is an array of globs, possibly negated.

`all` globs must ALL match ALL changed files. `some` globs must ALL match SOME changed files. This effectively gives us the AND and OR semantics I original described, but gives you finer-grain control over whether the globs are matches against all or some of the files.

Exclusions can now be represented as:
```
label:
- some: ["/example/**/*", "!/example/foo/*"]
```

However, you can also reject a label if a file exists with:
```
label:
- some: ["/example/**/*"]
   all: ["!/example/foo"]
```

One thing I did not do is add support for single elements without array or list syntax. I didn't want to complicate type checking, but I can add support if you think it's necessary.

ofek:
@jalaziz Great job! Hopefully I can get rid of our fork soon https://datadoghq.dev/integrations-core/meta/ci/#fork

ofek:
@dakale @pjquirk Can we please get another look at this?

dakale:
This looks pretty good so far. I see that the there are still README changes that are now outdated and might be misleading. Ideally, outside of full, up to date documentation, it would be best to at least avoid checking in outdated docs

Beyond that my only nit is: How do you feel about "any" instead of "some"? To me, "some" usually means > 1 whereas "any" means > 0?

jalaziz:
> This looks pretty good so far. I see that the there are still README changes that are now outdated and might be misleading. Ideally, outside of full, up to date documentation, it would be best to at least avoid checking in outdated docs
> 
> Beyond that my only nit is: How do you feel about "any" instead of "some"? To me, "some" usually means > 1 whereas "any" means > 0? 

Yeah, I'll be updating the docs, just wanted to make sure you're ok with the approach before adding docs and examples.

I'm definitely ok with any. I just chose some because of the JavaScript Array method name, but I much prefer any/all.

Will update the name and docs.

jalaziz:
@dakale everything should be updated.

I can squash once everything I get the final approval.

ericcornelissen:
Regarding the docs, I would expect the ["Basic Examples"](https://github.com/jalaziz/labeler/tree/exclusions#basic-examples) to come first, before the ["Match Object"](https://github.com/jalaziz/labeler/tree/exclusions#match-object) documentation, as the latter is more advanced.

jalaziz:
> Regarding the docs, I would expect the ["Basic Examples"](https://github.com/jalaziz/labeler/tree/exclusions#basic-examples) to come first, before the ["Match Object"](https://github.com/jalaziz/labeler/tree/exclusions#match-object) documentation, as the latter is more advanced.

@ericcornelissen I chose to place it before ["Basic Examples"](https://github.com/jalaziz/labeler/tree/exclusions#basic-examples) because it describes the "theory" of the examples. Happy to move it after the examples though.

dakale:
@jalaziz Im fine with merging this now unless you want to add anything else. No need to squash since I can just do squash+merge anyways. 

Ultimately, I think Ill need to go back and reorganize the docs anyways so they dont need to perfect here. I dont think the README is the best place to have extensive docs beyond a few small examples so ill need to measure the right amount to include here, and add some details around which version of labeler this will even be available in. Since the version of the README containing these samples is on the master branch, and that corresponds to the version of the action that has the described functionality, it should be fine. Although it would be ideal to describe it in terms of versions too.

Ill need to create a tag and release for this (will probably be v3), until then to be able to consume these changes users will need to use labeler@master, or using the sha of this merge commit (or newer).

ofek:
Yay! 🎉

jalaziz:
@dakale I have nothing to add if it all looks good to you.

akutz:
Mentioned in https://github.com/vmware-tanzu/vm-operator/pull/50#pullrequestreview-1274283991 (this was not added automatically for some reason)

PR #26, opened 2019-09-26
Re-worded readme with clearer examples

csi-lk:
Great package, just started using it with one of my projects, got a little confused with the examples thought I'd do some small re-wording of the readme

Let me know what you think

* Reformatted readme into headings with some basic rewording
* Added basic and common examples
Resolves #24

Issue #24, opened 2019-09-21
Add "any file with ending..." to Readme


Twonki:
Hey there, 
love labeler. 

I would like to see something alike:

```
Java:
- ./**/*.java
``` 

For the Readme.

I think this is a nice usecase e.g. for Dockerfiles

p942:
> Hey there,
> love labeler.
> 
> I would like to see something alike:
> 
> ```
> Java:
> - ./**/*.java
> ```
> 
> For the Readme.
> 
> I think this is a nice usecase e.g. for Dockerfiles

Hi na putanginanyo

dbaileychess:
I don't think the updated docs added a good example that was cited in the original issue. I just want to add a label to any PR that touches a .java file, no matter where it is located in the repo. 

It is necessary to specify this: `./**/*.java` which is not that intuitive. It complains when you just do `**/*.java`. So adding this specific example would be useful.

Issue #10, opened 2019-08-23
how to match path under any parent directory


craighurley:
Hello,

I have the following example folder structure

```
app/
└── subdir
    └── config
        └── subdir
            └── test.yaml
```

My workflow action looks like this:

```
name: Labeler
on: [pull_request]

jobs:
  label:

    runs-on: ubuntu-latest

    steps:
    - uses: actions/labeler@v2
      with:
        repo-token: "${{ secrets.GITHUB_TOKEN }}"
```

My goal is to be able to label PRs when certain files are found under certain sub-directories with the caveat that the target sub-directories may appear under _any_ parent directory in the repo.

This `labeler.yaml` works and results in a label being applied, but I have to specifically name the parent directory.

```
config:
- app/**/config/**/*.yaml
```

This `labeler.yaml` does not work/no label is applied:

```
config:
- ./**/config/**/*.yaml
```

These 2 variants of `labeler.yaml` fail with a similar error:

```
config:
- **/**/config/**/*.yaml
```

```
config:
- */**/config/**/*.yaml
```

... where the error is:

```
##[error]YAMLException: unidentified alias "*/config/**/*.yaml" at line 5, column 22:
    - **/config/**/*.yaml
                         ^
##[error]unidentified alias "*/config/**/*.yaml" at line 5, column 22:
    - **/config/**/*.yaml
                         ^
##[error]Node run failed with exit code 1 
```

Can someone advise if what I am trying to do is possible, if so, what should the glob pattern look like?

Thanks

craighurley:
Answering my own question...

Surround the pattern in single quotes so `labeler.yaml` looks like this:

```
config:
- '**/**/config/**/*.yaml'
```

PR #33, opened 2019-10-12
Add an example for a regex matching filename endings

jotaro-sama:
#24 asked how to do this but #26 didn't really answer the question, despite adding many useful examples. Making some tests starting from the intuition in #10, I came up with an example that actually matches all the files ending with a certain sequence in the whole repo.


MaksimZhukov:
Hello @jotaro-sama! Thank you for the contribution!
I am closing this PR since we've already added such example in scope of this [pull request](https://github.com/actions/labeler/pull/285).

Issue #28, opened 2019-10-02
labeler.yml example to match all repo files


lucascosti:
I'm trying to set up a rule to apply a `needs triage` label for _any_ file change in a repo, but I can't get it to work without manually specifying at least one folder in the path.

Here are some combinations I've tried without success:

```
needs triage:
  - '*'
```

```
needs triage:
  - '*'
  - '**/*'
```

```
needs triage:
  - ./*
  - ./**/*
```

```
needs triage:
  - '*'
```

It works for changes in a single folder if I provide at least part of a path, e.g.:

```
needs triage:
  - .github/*
```

Any help would be awesome!

cpanato:
same for me

azu:
similar issue? https://github.com/isaacs/minimatch/issues/30

Alfalfamale:
This worked for me:
```yml
needs triage:
  - '**'      # index.php | src/main.php
  - '.*'      # .gitignore
  - '.*/**'   # .github/workflows/label.yml
```

the `./` at the start doesn't do anything for me, it doesn't work

GMNGeoffrey:
Thanks I was trying to do the same thing and was confused why the "./*" syntax in the README wasn't working. It would be great if someone from the team addressed the broken documentation.

MaksimZhukov:
Hello everyone! 
Thank you @Robvdhout  for the provided solution! 
The confusing syntax (`./`) has been removed from the README. 
I’m closing the issue. Please contact us if you have any concerns.

PR #40, opened 2019-11-11
Adding example to match any file change

Robvdhout:
It is not clear how to match all files. This example will explain how to match any file.

has been tested with   
```yml
  - '**'      # index.php | src/main.php
  - '.*'      # .gitignore
  - '.*/**'   # .github/workflows/label.yml
```

The `./` at the start actually doesn't work

fixes https://github.com/actions/labeler/issues/28

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,12 @@ label2: example2/*
 repo:
   - ./*
   
+# Add 'triage' label to any file change
+triage:
+  - '**'
+  - '.*'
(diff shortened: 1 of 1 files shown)


carlislefox:
This works great, I think you guys should get this added the existing example for "all code" doesn't work and made me think the action wasn't working.

I'd go further and remove the ./ example altogether as I couldn't get it to pick up anything.

Issue #42, opened 2019-12-03
Unable to get all file changes.


Co-qn:
Unable to get all file changes.

Requests that return multiple items will be paginated to 30 items by default. 
c.f. https://developer.github.com/v3/#pagination

PR #43, opened 2019-12-03
Use pagination api to get all changes

Co-qn:
I tried to fix it. #42


ofek:
This works 😄 https://github.com/DataDog/integrations-core/pull/5384

dakale:
Thanks for this and sorry for the delay... the maintainership of this action has moved around a bit. I merged these changes via https://github.com/actions/labeler/pull/96

PR #44, opened 2019-12-16
Add support for regex globs and labels

alexkappa:
Adds support for regex globs and ability to replace matched groups in the label. 

I've found myself [repeating the same pattern](https://github.com/alexkappa/terraform-provider-auth0/blob/3511a3de133827d2380aa6ff3d8bf0bc16231deb/.github/labeler.yml) in order to add labels per resource, and thought it might be useful to support regex to allow using a match to customize the label.

Example:

```yml
api/$1:
  - 'api\\/resource_(\\w*?)(_test)?.go'
```

Given the files `api/resource_foo.go`, `api/resource_foo_test.go`, `api/resource_bar.go`, will produce the labels: `api/foo` and `api/bar`.

In order to be able to test this, I had to extract the labelling function into its own file so I could import it on its own (without calling `run()`).

I hope you'll find this useful, in what is already an awesome tool 😃

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,13 @@ test:
   - src/**/*.spec.js
 ```
 
+#### RegExp Examples
+
+```yml
+resource/$1:
+  - package\\/resource_(\\w*?)(_test)?.go
+```
+
 ### Create Workflow
 
 Create a workflow (eg: `.github/workflows/labeler.yml` see [Creating a Workflow file](https://help.github.com/en/articles/configuring-a-workflow#creating-a-workflow-file)) to utilize the labeler action with content:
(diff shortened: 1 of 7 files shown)


ofek:
@dakale Hello! Are there plans to merge/implement this functionality? This is the last thing requiring us to use our fork.

dakale:
Hey @ofek 👋 This is a pretty significant change, and definitely adds a lot of power, but I think it might go beyond anything I personally want to add at this time

I understand this PR is really old (sorry!), we've changed maintainership of this action a few times, and in the process adopted a slightly different non-regex based model, focusing on simple combinations of globbed paths, any/all, and exclusions (`!`). Of course, you can express all that (and much more) with a regex, so I think this overall a fantastic idea, just not one that I see integrating for now

Hopefully you wont feel discouraged from continuing to develop and use a fork of this action. I dont anticipate any/many upstream changes that youll need to worry about, since we get a lot of core functionality though the toolkit libraries

Cheers

Issue #36, opened 2019-10-21
Labels are not added on a forked repository


CsatariGergely:
I have a fork of cntt-n/CNTT in CsatariGergely/CNTT where I would like to test the labeler action before I propose a pr to the upsream repo. I've applied the example from the documentation with some minor changes, but no labels are added to [my test pr-s](https://github.com/CsatariGergely/CNTT/pulls) and becouse of #23 I do not know why. Can you please help me to figure out if

- Is it something my  [labeller.yml](https://github.com/CsatariGergely/CNTT/blob/master/.github/labeler.yml)?
- Is it becouse my repo is forked? There are some statements about forked repos in #12 , but I do not really understand them. In my case I would like to apply the labels of the forked repo to pr of the forked repo based on actions running in the forked repo. I see no differenve form a forked and a not forked repo in this sense. 
- Or what is it?

Thanks

ekohl:
This is indeed a duplicate of https://github.com/actions/labeler/issues/12

>     * I see no differenve form a forked and a not forked repo in this sense.

The action runs from the org of the source branch with the permissions of that. Since the fork doesn't have permissions to apply a label, it fails.

CsatariGergely:
@ekohl #12 have now an comment about everything, so I'm not sure if these are the same. 

When a pr is created the action should run on the target repo and target branch not the source.

ekohl:
That's a limitation of Github Actions and AFAIK not something you can control. My guess is that they chose to do this because of security reasons.

damianh:
This such an annoying limitiation. The whole github premise is that when creating a fork, even from private repo's in org, the fork goes to your personal account. It's not even possible to create a fork within the same user or org. This otherwise very useful action is useless for the standard operating proceedure. _sigh_

Ecco:
Yes, this is terrible. It's even been reported on this repository as early as #12 , even though this is not specific to this repository.

Actually I took the time to write [a comprehensive description of this problem](https://github.community/t5/GitHub-Actions/GitHub-actions-are-severely-limited-on-PRs/m-p/54669#M9249) on GitHub Community to explain the whole problem. Feel free to upvote it to raise awareness 😄

ylemkimon:
**tl;dr, change**
```yaml
on:
- pull_request
```
**to**
```yaml
on:
- pull_request_target
```

---

[...]

Issue #12, opened 2019-08-26
starter-workflow template gives `Resource not accessible by integration`


mahdipourziaei-binary:
Hi there,

I've implemented the exact Labeler workflow as the [starter-workflow template](https://github.com/actions/starter-workflows/blob/master/automation/label.yml) in the dir path `.github/workflows/label.yml`.

I have also added the `.github/labeler.yml` file with the configuration below:

```
Trader: packages/trader/**/*

Bot: packages/bot/**/*

Core: packages/core/**/*

Components: packages/components/**/*

Shared: packages/shared/**/*
```

The action is recognised and runs on PR, however the Labeler action resolves with the following error:

```
##[error]HttpError: Resource not accessible by integration
##[error]Resource not accessible by integration
##[error]Node run failed with exit code 1
```

Googling seems to relate that error with invalid access to the repo by the action app, so I thought I'd create an issue and disable the action for now, but would be happy to know if there's something I have missed out that will be able to solve this issue for me.

Thanks.

mahdipourziaei-binary:
Hey @damccorm, thanks for the response.
Yes, sure. The repo is over at [github.com/binary-com/deriv-app](https://github.com/binary-com/deriv-app). Although due to the failing action, we have currently disabled it, but you can find the [PR that enabled Labeler here](https://github.com/binary-com/deriv-app/pull/311), and an example of a [failed run here](https://github.com/binary-com/deriv-app/runs/203020900).

squidsoup:
I just submitted a support ticket about this, but probably should have checked here first! thanks @damccorm

mahdipourziaei-binary:
Thanks for investigating, @damccorm. And the update seems like a good alternative for my use at least. Others can reopen this if necessary, but I'll close it as it's expected behavior.

Just pinging you, @squidsoup to loop you in.

damccorm:
I spoke too soon. Switching this to cron by itself won't work because we [assume its going to be run on a PR](https://github.com/actions/labeler/blob/375fca6b7835f3dd43e68e699959851f8ab68102/src/main.ts#L13). Trying to figure out what makes the most sense here. Option 1 is to just not add labels to forks (but we shouldn't throw like we do now regardless). Option 2 is to update it to filter through all pull requests.

In theory I like option 2, but we need to be careful or we'll get rate limited - along those lines, we need a way of skipping PRs that we've already processed - maybe we could add a "triaged" label or something, but that's kind of ugly.

[...]

PR #49, opened 2020-02-01
Add support to schedule.

mshima:
Workaround to https://github.com/actions/labeler/issues/12 and https://github.com/actions/labeler/issues/36

When running from scheduled task, it will get every PR and process each one that don’t have tags yet.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -7,6 +7,11 @@ inputs:
   configuration-path:
     description: 'The path for the label configurations'
     default: '.github/labeler.yml'
+  not-found-label:
+    description: 'The label to apply when no matches were found'
+  operations-per-run:
+    description: 'The maximum number of operations per run, used to control rate limiting'
+    default: 30
 runs:
   using: 'node12'
   main: 'lib/main.js'
--- a/src/main.ts
+++ b/src/main.ts
@@ -3,19 +3,68 @@ import * as github from '@actions/github';
 import * as yaml from 'js-yaml';
 import {Minimatch} from 'minimatch';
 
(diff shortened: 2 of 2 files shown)


src/main.ts
```diff
@@ -7,15 +7,65 @@ async function run() {
...
 
+      if (await processPR(client, pr.number, configPath, notFoundLabel)) {
+        operationsLeft -= 1;
+      }
+    };
+  } catch (error) {
+    core.error(error);
+    core.setFailed(error.message);
+  }
+  return;
```
zekth: Isn't `async function run()` return a `Promise<void>` ? the 2 return statements change it.

mshima: L39 is return for filter

zekth: My bad for the filter :)

mshima: Done.

Removed catch from processPR too, otherwise a fail to label will be only a warning.

action.yml
```diff
@@ -7,6 +7,11 @@ inputs:
   configuration-path:
     description: 'The path for the label configurations'
     default: '.github/labeler.yml'
+  not-found-label:
+    description: 'The label to apply when no matches were found'
+  operations-per-run:
+    description: 'The maximum number of operations per run, used to control rate limiting'
+    default: 30
```
qaisjp: I don't think this is the correct way to implement rate limiting. Doesn't the Actions library handle rate limiting for you?

mshima: Got from another official action. I think it’s actions/stale

PR #50, opened 2020-02-02
Explain this action fails for fork/PR model

### LINKED ISSUE #12: starter-workflow template gives `Resource not accessible by integration`

yyyc514:
Resolves #12 via documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,8 @@
 
 Pull request labeler triages PRs based on the paths that are modified in the PR.
 
+**Caveat:** This only works when the PRs come from inside same account/organization. (if the PR is from a forked repo, like is commmon in many OSS projects, this action will not work due to lack of permissions).
+
 ## Usage
 
 ### Create `.github/labeler.yml`


ethomson:
Thanks @yyyc514, I'll include this with some minor tweaks.  Including it in #51

Ecco:
This is so bad. Here's a message I wrote to [raise awareness on this problem](https://github.community/t5/GitHub-Actions/GitHub-actions-are-severely-limited-on-PRs/m-p/54669#M9249), which is badly limiting the usefulness of Actions.

PR #52, opened 2020-03-04
Fix Code Block Type in Readme

thebeanogamer:
The final example in the README.md has not had a language defined, so does not have syntax highlighting. This PR resolves that by setting it to `YML`, the same as the other code blocks.


pje:
Huge fan of syntax highlighting. Thanks for the contribution!

PR #62, opened 2020-04-10
Prettier 2.0.4

singingwolfboy:
This upgrades us to the latest version of [Prettier](https://prettier.io/). See [the blog post introducing version 2.0](https://prettier.io/blog/2020/03/21/2.0.0.html).

I also switched to pinning an exact version of Prettier, [as suggested in the install docs](https://prettier.io/docs/en/install.html).

### DIFF
--- a/__tests__/main.test.ts
+++ b/__tests__/main.test.ts
@@ -1,4 +1,3 @@
-describe('TODO - Add a test suite', () => {
-  it('TODO - Add a test', async () => {
-  });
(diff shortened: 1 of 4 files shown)


pje:
Thanks for the contribution, @singingwolfboy!

I'm closing this PR (with nothing but appreciation in my heart) because it's been superseded by #151. 

> I also switched to pinning an exact version of Prettier, as suggested in the install docs.

Since we commit the `package-lock.json`, the version there is what counts.

PR #63, opened 2020-04-19
paginate over all changed files and sync and add "sync-labels" option to enable removing labels

### LINKED ISSUE #14: Remove labels for reverted changes
Reproducing:

1. Change something in a file in some PR.
2. A label will be added to the PR that matches the location of the changed file.
3. Revert the changes on the changed file. (i.e. the PR does not contain changes in such file).
4. The label will still be there on the PR. But it should be removed, since the PR does not affect such file anymore.

kevinphelps:
**Build process**
- run fomatter
  - This might be problem in my environment, but it looks like the code on `master` is not formatted via the `npm run format` command. I see the same changes in other open PRs so I don't think this is isolated to just me.
- generate dist/index.js in a deterministic manner
  - When I run `npm run build` it adds webpack-looking wrapper code to `dist/index.js` every time. It changes the js file even in main.ts is not changed. I changed the build script to `ncc build ./src/main.ts` to compile the TypeScript source directly and it seems to work better.
  - It is possible that this is a problem in my environment, but if it is, I can't seem to figure it out. Like the formatter issue, other open PRs seem to have unrelated changes in their `dist/index.js` as well so this also does not seem isolated to just me.

**Fixes**
- paginate over all changed files
  - This commit is pretty self-explanatory. This is the same source changes as #43, but I have updated `dist/index.js` also.

**Features**
- add "sync-labels" options to enable removing labels
  - When `sync-labels: true` is given, the action will remove any labels when no matching files are modified. This closes #14.

### DIFF
--- a/__tests__/main.test.ts
+++ b/__tests__/main.test.ts
@@ -1,4 +1,3 @@
-describe('TODO - Add a test suite', () => {
-  it('TODO - Add a test', async () => {
-  });
+describe("TODO - Add a test suite", () => {
+  it("TODO - Add a test", async () => {});
 });
--- a/action.yml
+++ b/action.yml
@@ -7,6 +7,10 @@ inputs:
   configuration-path:
     description: 'The path for the label configurations'
     default: '.github/labeler.yml'
+  sync-labels:
+    description: 'Whether or not to remove labels when matching files are reverted'
+    default: false
+    required: false
 runs:
   using: 'node12'
   main: 'dist/index.js'
--- a/dist/index.js
+++ b/dist/index.js
@@ -3545,12 +3545,13 @@ function getPrNumber() {
 }
 function getChangedFiles(client, prNumber) {
     return __awaiter(this, void 0, void 0, function* () {
-        const listFilesResponse = yield client.pulls.listFiles({
+        const listFilesOptions = client.pulls.listFiles.endpoint.merge({
             owner: github.context.repo.owner,
             repo: github.context.repo.repo,
             pull_number: prNumber
         });
-        const changedFiles = listFilesResponse.data.map(f => f.filename);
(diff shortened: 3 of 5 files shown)


ericcornelissen:
Nice work on the "sync-labels" feature :+1:

Perhaps it is an idea to limit this PR to just that feature to increase the likelihood of it being merged? Pagination is already implemented in #43 and should probably be merged there. Doing the formatting in this PR makes it difficult to review the meaningful changes in this PR (and causes unnecessary merge conflicts).

kevinphelps:
I can rebase this later and see if the formatting is fixed on `master`. I only made those changes in this PR because I could not get a passing build otherwise.

ericcornelissen:
> I can rebase this later and see if the formatting is fixed on master. I only made those changes in this PR because I could not get a passing build otherwise.

Makes sense. The formatting is not yet fixed on master, but it seems you will no longer get a failing build as a result of incorrect formatting.

kevinphelps:
The other issue is that `npm run build` with `master` checked out changes the `dist/index.js` build even though none the TypeScript is changed. I am unsure how to handle this.

ericcornelissen:
> The other issue is that npm run build with master checked out changes the dist/index.js build even though none the TypeScript is changed. I am unsure how to handle this.

I can't help you with that, could be due to a variety of reasons. But I don't think it matters much. Merge conflicts in the dist/index.js file are straightforward to resolve by the maintainers of this action and don't affect reviewing the changes, so I wouldn't worry about it.

dakale:
Thanks for this, and sorry for the huge delay. The maintainership of this action has changed a few times so some older items like this fell off the radar

I merged your changes via https://github.com/actions/labeler/pull/96 to make it a little easier / faster as Im going through and cleaning up a bit

PR #69, opened 2020-05-08
Bumping actions/core and actions/github dependencies for GHES

pjquirk:
This adds support for running against GHES by using [v2.2.0](https://github.com/actions/toolkit/blob/master/packages/github/RELEASES.md#220) of `@actions/github`.  I also bumped `@actions/core` to 1.2.4 to fix error handling in the case an `Error` type is thrown rather than a `string`.

The build/pack steps seemed unreliable, so I updated the Typscript config to build to `./lib`, and `ncc` will then build out of that (this matches other actions like [actions/stale](https://github.com/actions/stale/blob/master/package.json#L6)).

Tested by pushing this version of the action to an "actions" org on the GHES instance.  I also needed to set `NODE_TLS_REJECT_UNAUTHORIZED = 0` temporarily in the action to get around my self-signed (or otherwise not trusted) server cert.


lib/main.js
```diff
@@ -0,0 +1,135 @@
+"use strict";
```
ericsciple, line 1: git ignore?

ericsciple: may need to `git rm`

dakale:
What is the benefit of changing from dist/index to lib/index? Not sure I understand the point, or why the lib/ file is so much smaller than the dist/ one...

pjquirk:
@dakale It's mainly to distinguish the outputs of `tsc` (lib) and the outputs of `ncc` (dist).  It doesn't make sense to put them both into the same folder, like it was before.  `dist` should have one file in it, `index.js`.

dakale:
Thanks, that makes sense, as does ignoring checking in lib/

dakale:
(I can merge if you dont have permission)

taehwanno:
It seems that `dist/index.js` was not built with the correct dependency code. If you look at the [`v1.2.4` release notes of @actions/core](https://github.com/actions/toolkit/blob/1cc56db0ff/packages/core/RELEASES.md#124), there are changes to [non-string command inputs](https://github.com/actions/toolkit/pull/405), but there is no related content in the built `dist/index.js`. It is related to #88 and needs to be published with a new version. Maybe #70 can be solved.

PR #72, opened 2020-05-14
Bump actions/github to 2.2.0 to support GHES

pjquirk:
This adds support for running against GHES by using [v2.2.0](https://github.com/actions/toolkit/blob/master/packages/github/RELEASES.md#220) of `@actions/github`. The details of connecting to a GHES instance or github.com are handled internally in the client, so no code changes in this action are necessary.

Tested by pushing this version of the action to an "actions" org on the GHES instance.  I also needed to set `NODE_TLS_REJECT_UNAUTHORIZED = 0` temporarily in the action to get around my self-signed (or otherwise not trusted) server cert.


pjquirk:
I had to make a few small changes to account for the octokit upgrade.  Changes are [here](https://github.com/actions/labeler/blob/c3d2ae9c536660c2ca23b20259d9abc4c722d9dd/src/main.ts#L99), since the files changed tab is a nightmare.

pjquirk:
Tag is moved forward: https://github.com/actions/labeler/tree/v2

Amoon-123:
on : push
 jobs : 
  test : 
    matrix : platform : [ ubuntu-latest، macos-latest، windows-latest]
     التشغيل : $ {matrix.platform}}
     الخطوات : 
    - الاستخدامات : Actions / checkout @ v3
      
        
    - يستخدم : Actions / setup-node @ v3
       مع : 
        إصدار العقدة : 16
    - تشغيل : npm install-ci-test
    - الاستخدامات :

PR #75, opened 2020-06-10
Fix "any" matching when not all globs match

jalaziz:
The previous logic had a bug where the "any" pattern list could match
against changed files even when not all globs matched a single changed
file. The bug arose individual globs in the "any" list were tested
against all changed files individually. Therefore, as long as at least
one changed file matched an individual glob in the list, a successful
match would be found.

The correct behavior is to match all the globs in the list against each
individual file. This ensures that is possible to define exlcusions
correctly and matched the documented behavior.

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -4207,6 +4207,9 @@ function toMatchConfig(config) {
     }
     return config;
 }
+function printPattern(matcher) {
+    return (matcher.negate ? "!" : "") + matcher.pattern;
+}
 function checkGlobs(changedFiles, globs) {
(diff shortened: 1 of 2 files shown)


jalaziz:
@dakale Still testing this, but discovered a bug in the previous logic.

jalaziz:
Ran some tests against a test repo I have and I believe this logic works and fixed the issue.

jalaziz:
@dakale Just wanted to ping on this since master is currently buggy.

jalaziz:
Thanks!

PR #80, opened 2020-06-21
Two minor fixes to the README

ericcornelissen:
- ~~Add syntax highlighting to the "Create Workflow" example YAML~~ See #52
- Update examples to use consistent indentation style (~~I choose to use indentation as much as possible, we can also go for the opposite~~)
- Fixed a typo ("invidual" -> "individual")

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,7 +81,7 @@ frontend:
 
 Create a workflow (eg: `.github/workflows/labeler.yml` see [Creating a Workflow file](https://help.github.com/en/articles/configuring-a-workflow#creating-a-workflow-file)) to utilize the labeler action with content:
 
-```
+```yml
 name: "Pull Request Labeler"
 on:
 - pull_request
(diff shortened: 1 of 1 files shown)


dakale:
Thanks for this! Personally, I prefer unindented (nit picky) but if it is not consistent to that style it should be updated. Typo fix is definitely welcome as is, of course

ericcornelissen:
> Personally, I prefer unindented (nit picky) but if it is not consistent to that style it should be updated.

I will update so that all instances are unindented instead of indented, I have no strong preference other than it being consistent :smile:

dakale (approved):
LGTM 👍 Thanks!

ericcornelissen:
Should this PR have targetted the `main` branch instead of `master`? Both issues fixed in this PR are still present in the README on `main` :sweat_smile: Should I open a new PR with the exact same changes for that?

jsoref:
@dakale: ping. (Just noticed this divergence while investigating master/main use in https://github.com/actions)

pje:
@ericcornelissen 👋 https://github.com/actions/labeler/pull/146

PR #82, opened 2020-06-27
Pin README example to master

TimothyGu:
The current README describes features that are only in master, with many of the examples not working with `@v2` or `@v3-preview`. Pin the action version in the sample workflow file to `master`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,7 +90,7 @@ jobs:
   triage:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


dakale:
Thanks for bringing this up. In the time since, we've switched to using `main` as the default branch, so Ive updated that and included the spirit of your change in: https://github.com/actions/labeler/commit/bab130482f3d177a5ecf31d8c2cea14fb04f8855

PR #83, opened 2020-07-03
Update .gitattributes file

ericcornelissen:
With two configurations:

1. `* text=auto`: normalizes line endings for all files in the repo.
2. `FILE -diff`: hides the diff of these files by default. I configured this for _dist/index.js_ and _package-lock.json_ as they are not (intended to be) human readable nor edited by humans. Hence, seeing their diff is mostly useless.


ericcornelissen:
Should this PR be updated to target the `main`  branch (instead of `master`)?

pje:
Thank you, @ericcornelissen!

PR #90, opened 2020-08-08
Update README to use `pull_request_target` event

ylemkimon:
GitHub [has introduced](https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/) a new event type: `pull_request_target`, which allows to run workflows from base branch and pass a token with write permission.

> In order to solve this, we’ve added a new `pull_request_target` event, which behaves in an almost identical way to the `pull_request` event with the same set of filters and payload. However, instead of running against the workflow and code from the merge commit, the event runs against the workflow and code from the base of the pull request. This means the workflow is running from a trusted source and is given access to a read/write token as well as secrets enabling the maintainer to safely comment on or label a pull request. This event can be used in combination with the private repository settings as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,8 +2,6 @@
 
 Pull request labeler triages PRs based on the paths that are modified in the PR.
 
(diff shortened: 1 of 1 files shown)


dakale:
LGTM, thanks for this!

PR #96, opened 2020-09-08
Paginate and remove old labels

### LINKED ISSUE #14: Remove labels for reverted changes

### LINKED ISSUE #42: Unable to get all file changes.

dakale:
Big thank you to @kevinphelps for https://github.com/actions/labeler/pull/63 (closes #14) and @Co-qn for https://github.com/actions/labeler/pull/43 (closes #42)

I just took those changes verbatim and rebased them and re-created this for simplicity. In the future I will *ideally continue to merge existing PRs rather than create new ones* but given their age I felt it reasonable to recreate them here so they'd be rebased, I could generate an up to date build, and run the formatting

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -14,6 +14,7 @@ async function run() {
   try {
     const token = core.getInput('repo-token', {required: true});
     const configPath = core.getInput('configuration-path', {required: true});
+    const syncLabels = !!core.getInput("sync-labels", { required: false });
 
     const prNumber = getPrNumber();
     if (!prNumber) {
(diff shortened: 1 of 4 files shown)


src/main.ts
```diff
@@ -12,17 +12,24 @@ type StringOrMatchConfig = string | MatchConfig;
 
 async function run() {
   try {
-    const token = core.getInput('repo-token', {required: true});
-    const configPath = core.getInput('configuration-path', {required: true});
+    const token = core.getInput("repo-token", { required: true });
+    const configPath = core.getInput("configuration-path", { required: true });
+    const syncLabels = !!core.getInput("sync-labels", { required: false });
```
stevecotton, line 17: sync-labels still triggering when set to `false` (bug #104), or when unset.

I think the bug is on this line - getInput returns a string, and `!!"false"` is true, just like `!!"true"`.

kevinphelps: Doh! When I orginally wrote this code, I don't think I tested setting the input to `false`. You can just leave the input out to disable syncing as a workaround until this bug is fixed.

Issue #81, opened 2020-06-21
Contributors documentation is incorrect


ericcornelissen:
The [docs/contributors.md](https://github.com/actions/labeler/blob/efc1b29f09efe879241e3b11e710d114dc56ef1f/docs/contributors.md) file on the `main` branch is outdated:

- Husky is no longer used by this project
- The runtime dependencies in node_modules should no longer be committed
- The build output folder has been renamed from "lib" to "dist"

I'd be willing to write something up to update the existing text (e.g. based on [the Contributing Guidelines of `actions/cache`](https://github.com/actions/cache/blob/eed9cfe64d00dd64cbb36bec915d41dd6fab9f6c/CONTRIBUTING.md)) but as a outsider I'm not sure if I'm the right person to do that.

dakale:
Thanks for bringing this up. Its definitely an area that needs work. I think theres plenty of things anyone can contribute here. A lot of the contributing docs are just welcoming, encouraging the fork+pr model, etc. Thats something I can drop in as a general representation of how we at GitHub work with the community

There are also project specific details that might warrant including. Some of them you mentioned: what tools are used, what files are committed and how things are structured, and how to build and test (well, that process needs work too). Its fairly tricky because I think the discrepancies are coming from the fact that we migrated a lot of actions from husky->ncc, and have updated the general model that we recommend for building actions (checking in full out including node_modules vs something more webpack-y), but I made that switch for this action as a version change, for safety. So theres a v2 labeler being released from `releases/v2` as to not break existing users, and that has a different set of conventions than the v3+ version that Im releasing from `main` 

Its probably worth focusing most those development details that you mentioned on the new model, so I totally agree with the premise. I dont expect many contributors to have interest in maintaining the "legacy" branch, but that information should probably live *somewhere*, for future maintainers

Definitely something ill think about as time permits!

ericcornelissen:
> Thanks for bringing this up. Its definitely an area that needs work. I think theres plenty of things anyone can contribute here. [...]

In that case I will see if I can get a PR out with some basic improvements :smile: As an aside, any particular reason why this repository is not (or should not) using the more standard(?) `CONTRIBUTING_GUIDELINES.md` file at the root of the repo?


> [...] A lot of the contributing docs are just welcoming, encouraging the fork+pr model, etc. Thats something I can drop in as a general representation of how we at GitHub work with the community

[...]

PR #97, opened 2020-09-10
Draft contributors.md update

ericcornelissen:
This is a proposed updated to [the contributors.md document](https://github.com/actions/labeler/blob/bab130482f3d177a5ecf31d8c2cea14fb04f8855/docs/contributors.md), addressing #81, inspired by [the contributing guidelines of the actions/cache project](https://github.com/actions/cache/blob/eed9cfe64d00dd64cbb36bec915d41dd6fab9f6c/CONTRIBUTING.md). There are some differences, most notable the inclusion of "Development Details" which explains and motivates the development tools (to an extend).

The changes here are completely open for discussion. I'm open to suggestions and feedback and would be happy to update this PR accordingly (or have it updated by the maintainers directly).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -19,12 +19,6 @@ Here are a few things you can do that will increase the likelihood of your pull 
 - Keep your change as focused as possible. If there are multiple changes you would like to make that are not dependent upon each other, consider submitting them as separate pull requests.
 - Write [good commit messages](http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html).
 
-## Development Details
-
-This project is written in [TypeScript](https://www.typescriptlang.org/), a typed variant of JavaScript, and we use [Prettier](https://prettier.io/) to get a consistent code style.
-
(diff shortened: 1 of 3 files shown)


dakale:
This looks great so far! I think I agree with what you said in https://github.com/actions/labeler/issues/81 about:

> As an aside, any particular reason why this repository is not (or should not) using the more standard(?) CONTRIBUTING_GUIDELINES.md file at the root of the repo?

I think we should split this out into two files: `CONTRIBUTING.md` containing the conribution guidelines, and a second file `docs/development.md` that will serve as a development handbook, starting with the ones in this doc. That way we have a clear split between between "we welcome contributions (of any kind) through issues/prs" and if those contributions are code, "heres how to develop/test the actual code". That way if the dev handbook grows, it doesnt crowd out the simple, approachable CONTRIBUTING guidelines

ericcornelissen:
That's an interesting way of structuring things. I do like it, so I updated the PR. my only concern is that the development handbook for this project won't be very large and could therefore easily be included in `CONTRIBUTING.md`... But it's up to the maintainers of the project 😄

dakale:
Looks great, thanks! 

> my only concern is that the development handbook for this project won't be very large and could therefore easily be included in `CONTRIBUTING.md`... 

Yea, that makes sense. My hope is that it will grow over time though, as theres not much there right now

PR #116, opened 2020-11-27
 Suggestion: auto-update dist/index.js on main branch

ericcornelissen:
This is a suggestion to change the way contributing to this project works. Currently, when a PR is submitted the submitter has to run `npm run build` and include the changes to `dist/index.js` in the PR. With the changes in this PR, including the changes to `dist/index.js` is no longer necessary (*). This should avoid a bunch of merge conflicts with regards to the `dist/index.js` file - even though they're simple to fix it is annoying you have to. I updated the [contributing guidelines](https://github.com/actions/labeler/blob/9794b1493b6f1fa7b006c5f8635a19c76c98be95/CONTRIBUTING.md) accordingly, ~~removing~~ editing the fourth step of "Submitting a Pull Request".

<sup>(*): It is still possible though, and if a PR is merged with updates to lib/ included, the changes in this PR won't have any effect.</sup>


pje:
@ericcornelissen I broadly agree, and I sketched out the implementation here: https://github.com/actions/labeler/pull/155

Comments/improvements welcome!

ericcornelissen:
While similar, @pje, it seems #155 does something slightly different, so I'm not sure what you're getting at with your comment :thinking: 

That being said, #155 looks great!

pje:
@ericcornelissen right, what I meant is: that PR is the same principle ("automatically commit machine-generated changes in response to changes to a specific file"), just with different targets. 👍

Issue #107, opened 2020-10-19
Labeler is not labeling PRs


Kitzunu:
Hey! 

I am using this action on an open-source repository with these settings:

```yml


DB:
  - data/**/*.sql

CORE:
  - any: ['src/**/*.cpp', 'src/**/*.h', '!src/server/scripts/*']

Script:
  - src/server/scripts/**/*.cpp
  - src/server/scripts/**/*.h

Documentation:
  - ./*.md

Bash:
  - ./*.sh

Cmake:
  - ./*.cmake

Workflow:
  - .github/workflows/*
```
```yml
name: "Pull Request Labeler"
on: [pull_request]

jobs:
  triage:
    runs-on: ubuntu-20.04
    steps:
    - uses: actions/labeler@main
      with:
        repo-token: "${{ secrets.GITHUB_TOKEN }}"
        configuration-path: .github/labeler.yml
        sync-labels: true
```

On the initial PR to implement this it did label on what was changed no problems, but now when I update PRs to latest master the action will run, however, it will not label anything.

Is there something in my settings that I have messed up for it to correctly label PRs?

Kitzunu:
Okay, so I tested a few other directories and it looks like the "Workflow", "Scripts" and "DB" labels works, however, none of the ./* paths or paths inside the *any* will get a label on them

mgan59:
Just came here, and believe my rule:

```yaml
docs:
  - ./*.md
  - docs/*.md
  - docs/**/*.md
```

having a similar issue.  Labels for things in `docs/` but for instance changes to the root `readme.md` don't get labeled.

kenbalto:
Has anyone found a fix for this?

Kitzunu:
Okay I changed to two conditions as this
```yml
  - any: ['src/**/*', '!src/server/scripts/*']
```
Now it seems to be working as it should

MaksimZhukov:
Hello everyone!
This issue has been open for a year without any activity, so I'm going to close it as stale.
Please feel free to contact us if the issue is still relevant to you and we will reopen it.
Thanks!

brysonjonesy:
> On the initial PR to implement this it did label on what was changed no problems, but now when I update PRs to latest master the action will run, however, it will not label anything.

Same thing is happening to us

Issue #46, opened 2020-01-11
Support relative path from the root.


tyankatsu0105:
# Describe
I created `CHANGELOG.md` in root.
I saw https://github.com/actions/labeler#common-examples :
```yml
repo:
  - ./*
```

Then, I try:
```yml
release:
  - ./CHANGELOG.md
```
But, it's failed.

```yml
release:
  - CHANGELOG.md
```
is succeed.

# Expect
Support relative path from the root.
```yml
release:
  - ./CHANGELOG.md
```

dogmatic69:
I've got a similar issue but I think it can

dogmatic69:
I've got a similar issue but I think relative paths would fix my issue.

I want to add a tag to every PR, so I've tried to add `./**` which is not possible due to missing the relative paths. `**` is not valid either 

Edit: Now figured out if I do `'**'` it will add the label to all PRs

NobbZ:
As "relative to the root" means the same as "absolute", have either of you tried `/` instead of `./`?

I do not use labeller currently because #12 and therefore can't test currently.

PR #133, opened 2021-03-15
Fix documentation for root file changes

GisoBartels:
might also resolve the confusion in #46 & #107


hanselabreu:
This checks out, I use this exact label on a repo and it didn't work until doing this change.

pje:
That is indeed confusing behavior based on the documentation. We appreciate the fix! Thanks!

PR #138, opened 2021-04-22
DOC: Example to use released version

pllim:
Released version is the one people should use, right?


lastthyme (approved):

pje:
> Released version is the one people should use, right?

I mean, `main` _should always_ be stable... 😅 

But seriously, yes, I think you're right—let's tell people to use the latest release tag.

PR #146, opened 2021-06-02
🧹 Merge commits from "master" into "main"

pje:
Awhile back we switched from `master` to `main`. Looks like one minor commit got added to `master` but not `main`.

After this I can just delete `master` so there's less confusion in the future.


pje:
https://github.com/github/c2c-actions-runtime/issues/1273

PR #148, opened 2021-06-02
🧹 Add some really rudimentary unit tests

pje:
## Goals

Add some basic tests!

## Non-Goals

This is not intended to be a comprehensive test suite; I'm just trying to get us started by adding tests for some of the most important functionality.

## How

This necessitated some minor refactoring. Previously, `main.ts` just directly invoked the main entrypoint function `run()`, which made it impossible to unit test the module without causing the side-effects of `run`. 

As a workaround I created  a new module `labeler.ts`, which just exports the functions we want to test without executing the main entrypoint, and simplified `main.ts` so that it only imports the entrypoint and executes it.

It's basically just a re-org. The diff makes it look way more complicated than it is.


pje:
https://github.com/github/c2c-actions-runtime/issues/1273

src/main.ts
```diff
@@ -1,261 +1,3 @@
...
-      client.issues.removeLabel({
-        owner: github.context.repo.owner,
-        repo: github.context.repo.repo,
-        issue_number: prNumber,
-        name: label
-      })
-    )
-  );
-}
+import { run } from "./labeler";
```
TingluoHuang, line 1: do we have to make this change? anything changed between `labeler.ts` and `main.ts`?

__tests__/labeler.test.ts
```diff
@@ -0,0 +1,29 @@
+import { checkGlobs } from '../src/labeler'
```
TingluoHuang, line 1: 🎉  we have tests... 😆

src/main.ts
TingluoHuang: i am fine with changing it, just want to make sure there are no functional changes since the diff is not really helpful. 😄

TingluoHuang (approved):

__tests__/labeler.test.ts
pje: 😆 🚀 🖥️

src/main.ts
pje: > anything changed between labeler.ts and main.ts?

No changes. I just moved everything from main into labeler to avoid invoking `run()` at the end of the file, so we can test it in isolation. The diff makes it look like a bigger change than it is :)

PR #152, opened 2021-06-07
Upgrade to typescript 4.3 and jest 27

pje:
This upgrades us to the latest versions of typescript and jest.


jest.config.js
```diff
@@ -3,7 +3,6 @@ module.exports = {
   moduleFileExtensions: ['js', 'ts'],
   testEnvironment: 'node',
   testMatch: ['**/*.test.ts'],
-  testRunner: 'jest-circus/runner',
```
pje, line 6: > As of Jest 27, jest-circus is the default test runner, so you do not have to install it to use it.

https://www.npmjs.com/package/jest-circus

thboop (approved):
LGTM

PR #153, opened 2021-06-07
Upgrade licensed to 3.0.1

pje:
The version we install during the Actions workflow (2.12.2) wasn't checking transitive dependencies in some cases when installed locally on macOS. Upgrading to 3.0.1 fixes that.


thboop (approved):
LGTM, little annoying the new version drops URLS though

PR #155, opened 2021-06-07
Workflow automatically updates license cache when package-lock.json changes

pje:
This branch is a spike. I'm not convinced it's the right approach, but I wanted to get my thoughts down formally and a Draft PR with working code seemed better than an abstract issue.

### Problem

Whenever we upgrade a dependency, our license cache becomes out of date and the build fails. This is confusing and requires manual intervention for a task that is ultimately Easy To Automate™.

### Approach

The changes in this branch define an Action that automatically regenerates the license caches in response to PRs with changes to `package-lock.json` and commits the result.

### Alternatives

1. A simple pre-commit hook could do the same. Downside: it'd require having a local installation of `licensed`.


.github/workflows/licensed_cache_update.yml
```diff
@@ -0,0 +1,27 @@
...
+    runs-on: ubuntu-latest
+    name: Check licenses
+    steps:
+      - uses: actions/checkout@v1
+      - uses: jonabc/setup-licensed@v1.0.2
+        with:
+          version: '2.x'
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+      - run: npm install --production
+      - uses: jonabc/licensed-ci@v1.5.3
```
thboop, line 25: Consider pinning to v1 unless you have a specific reason to do this

.github/workflows/licensed_check.yml
```diff
@@ -0,0 +1,19 @@
...
+  pull_request: {branches: main}
+
+jobs:
+  licensed_check:
+    needs: licensed_cache_update
+    runs-on: ubuntu-latest
+    name: Check licenses
+    steps:
+      - uses: actions/checkout@v2
+      - uses: jonabc/setup-licensed@v1.0.2
```
thboop, line 14: Same as above, consider pinning to v1

.github/workflows/licensed_cache_update.yml
```diff
@@ -0,0 +1,27 @@
+name: Licensed Cache Update
+
+on:
+  push:
+    branches:
+      - main
+      - licensed_cache_update
```
thboop, line 7: we should remove this before merging

.github/workflows/licensed_cache_update.yml
pje: Agree!

Issue #101, opened 2020-09-23
Issue of the label configuration file ".github/labeler.yml"


BrightRan:
Associated GitHub Community ticket: https://github.community/t/labeler-typeerror-glob-pattern-string-required/133467

According to the [README](https://github.com/actions/labeler#create-githublabeleryml) of this action, the following configurations in the label configuration file should work:
```yaml
label-1:
  - any: ['list', 'of', 'globs']
    all: ['list', 'of', 'globs']

label-2:
  - example1/*
  - example2/**/*
```

I tested with two label configuration files in my repository (see [here](https://github.com/TestWorkflowsRML/PR-Labeler/tree/test-PR/.github)).
* **.github/labeler-01.yml**
  ```yaml
  common:
    - any: ['./*', './src/**']

  workflow:
    - any: ['.github/workflows/**']
  ```
* **.github/labeler-02.yml**
  ```yaml
  repo:
    - './*'
    - './src/**'

  GHA:
    - '.github/**'
  ```

And my workflow looks like this (see [here](https://github.com/TestWorkflowsRML/PR-Labeler/actions/runs/268032638/workflow)):
```yaml
name: CI

on: pull_request

jobs:
  job1:
    runs-on: ubuntu-latest
    steps:         
      - name: PR Labeler
        uses: actions/labeler@2.2.0
        with:
          repo-token: ${{ secrets.GITHUB_TOKEN }}
          configuration-path: .github/labeler-01.yml

  job2:
    runs-on: ubuntu-latest
    steps:         
      - name: PR Labeler
        uses: actions/labeler@2.2.0
        with:
          repo-token: ${{ secrets.GITHUB_TOKEN }}
          configuration-path: .github/labeler-02.yml
```
I changed the following files:
* ./.github/workflows/ci.yml
* ./test.txt
* ./src/test.txt

However, finally only the **`GHA`** label was added successfully, the other three should also be added but not (see the [example PR](https://github.com/TestWorkflowsRML/PR-Labeler/pull/3)).

When using '**.github/labeler-01.yml**', the actions always fails with the error (see [job1](https://github.com/TestWorkflowsRML/PR-Labeler/runs/1152881395?check_suite_focus=true#step:2:2)):
```bash
##[error]TypeError: glob pattern string required
##[error]glob pattern string required
```

[...]

Issue #55, opened 2020-03-11
Support labeling based on the pull request/commit title and messages


XVilka:
To apply the label based on the pattern match in one of the fields in:
- PR title
- PR body
- PR commit title
- PR commit body

austince:
The commit-specific matching might be more work, as it would require fetching all the commits that are contained in the PR, but the PR title and body should be already available.

What about adding a config object `pr: { body?: StringOrMatchConfig, title?: StringOrMatchConfig }`?

Ex:
```yaml
# Add 'test' label to any change to *.spec.js files within the source dir
# or with a PR title that is in the form 'test:*'
# or with a PR body that contains the word 'test' but not 'bug'
test:
 - src/**/*.spec.js
 - pr:
     title:
       - 'test:*'
     body:
       - any: [ '*test*' ]
         all: [ '*bug*' ]
```

This could then be expanded to include the commits:

```yaml
# Add 'test' label to any change to *.spec.js files within the source dir
# or with a PR title that is in the form 'test:*'
# or with a PR body that contains the word 'test' but not 'bug'
# or has any commits with a title in the form 'test:*'
test:
 - src/**/*.spec.js
 - pr:
     title:
       - 'test:*'
     body:
       - any: [ '*test*' ]
         all: [ '*bug*' ]
- commits:
    title:
      - 'test:*'
```

felipefrancisco:
+1

silasdavis:
Can we also have the ability to specifically label based on git commit footers, e.g.

```
chore(foo): my fix

Risk: D
Urgency: 2

Signed-off-by: Silas Davis <silas.davis@monaxlabs.io>
```

This would play nicely with conventional commits, for example: https://www.conventionalcommits.org/en/v1.0.0/#specification

kalvarez2:
Hi, would it be acceptable to instead of the structured way, "simply" to add an optional prefix on the blob?

For example:

```
label1:
- "body:#potato"

```
Would add the label "label1" if the body contains the string #potato

Looking at the code, this would be simpler to add, and from the user experience point of view I think it works ok to?

Jaskowicz1:
I like that, would be extremely useful!

kalvarez2:
@MaksimZhukov should I make another for 5.0 branch? is there any hope that this will be merged to 4.0?
https://github.com/actions/labeler/pull/688

sschuberth:
[...]

Issue #54, opened 2020-03-06
Label PR based on base branch


hkdobrev:
Could the automatic labeler support more triggers than just the paths changed?

A very useful workflow for us would be set labels on the PR depending on the base branch of the PR. Depending on the Git branching workflow used in a project, this could have different use cases. For example, using the git-flow, a PR towards the development branch could be labeled as a feature and one towards master as a release.

If this feature request is implemented I imagine automatic labeling and its configuration could later be expanded to be based on branch patterns, author (if it's a robot for example) etc.

XVilka:
Also related https://github.com/actions/labeler/issues/55 - probably implementing these will share some code.

MaksimZhukov:
Hello everyone!
We released the new alpha version of the action that includes the implementation of this feature. For more information, see the [release notes](https://github.com/actions/labeler/releases/tag/v5.0.0-alpha.1). We would really appreciate it if you could try it out and provide us with [your feedback](https://github.com/actions/labeler/issues/575). 

I will keep this issue open until the stable version is released.
Thank you!

MaksimZhukov:
Hello everyone! 
We have released the [new major version of the action](https://github.com/actions/labeler/releases/tag/v5.0.0) that includes the implementation of this feature. Please read the [action documentation](https://github.com/actions/labeler/tree/main#pull-request-labeler) to find out how to adapt your configuration files and workflows for use with the new action version.

Thank you all for your patience!
Do not hesitate to contact us if you have any questions!

PR #203, opened 2021-08-07
Assigns labels based on branch names

joshdales:
Assign a label based on the branch name! 

Depending on your branch name you may want to apply different labels based on the purpose of the branch.

eg. say that you wanted to something like:
```yml
feature:
 - branch: feature/*
bug:
 - branch: */bug/*
```
It would apply the `feature` label to the branch to `feature/add-the-new-thing` branch, and the `bug` label to the `josh/bug/fix-the-new-thing` branch.

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -203,6 +203,18 @@ function checkAll(changedFiles, globs) {
     core.debug(`  "all" patterns matched all files`);
     return true;
 }
+function checkBranch(glob) {
+    const matcher = new minimatch_1.Minimatch(glob);
+    const branchName = github.context.ref;
+    core.debug(` checking "branch" pattern against ${branchName}`);
+    core.debug(`  - ${printPattern(matcher)}`);
+    if (!matcher.match(branchName)) {
+        core.debug(`   ${printPattern(matcher)} did not match`);
+        return false;
+    }
+    core.debug(`   "branch" pattern matched`);
+    return true;
+}
 function checkMatch(changedFiles, matchConfig) {
     if (matchConfig.all !== undefined) {
         if (!checkAll(changedFiles, matchConfig.all)) {
@@ -211,6 +223,11 @@ function checkMatch(changedFiles, matchConfig) {
     }
     if (matchConfig.any !== undefined) {
         if (!checkAny(changedFiles, matchConfig.any)) {
+            return false;
+        }
+    }
+    if (matchConfig.branch !== undefined) {
+        if (!checkBranch(matchConfig.branch)) {
             return false;
         }
     }
@@ -11709,7 +11726,7 @@ module.exports = eval("require")("encoding");
 /***/ ((module) => {
 
 "use strict";
-module.exports = require("assert");;
+module.exports = require("assert");
 
 /***/ }),
 
@@ -11717,7 +11734,7 @@ module.exports = require("assert");;
 /***/ ((module) => {
 
 "use strict";
-module.exports = require("events");;
+module.exports = require("events");
 
 /***/ }),
 
@@ -11725,7 +11742,7 @@ module.exports = require("events");;
 /***/ ((module) => {
 
 "use strict";
-module.exports = require("fs");;
+module.exports = require("fs");
 
 /***/ }),
 
@@ -11733,7 +11750,7 @@ module.exports = require("fs");;
 /***/ ((module) => {
 
 "use strict";
-module.exports = require("http");;
+module.exports = require("http");
 
 /***/ }),
 
@@ -11741,7 +11758,7 @@ module.exports = require("http");;
 /***/ ((module) => {
 
 "use strict";
-module.exports = require("https");;
+module.exports = require("https");
 
 /***/ }),
 
@@ -11749,7 +11766,7 @@ module.exports = require("https");;
 /***/ ((module) => {
 
 "use strict";
-module.exports = require("net");;
+module.exports = require("net");
 
 /***/ }),
 
@@ -11757,7 +11774,7 @@ module.exports = require("net");;
 /***/ ((module) => {
 
 "use strict";
-module.exports = require("os");;
+module.exports = require("os");
 
 /***/ }),
 
@@ -11765,7 +11782,7 @@ module.exports = require("os");;
 /***/ ((module) => {
 
 "use strict";
-module.exports = require("path");;
+module.exports = require("path");
 
 /***/ }),
 
@@ -11773,7 +11790,7 @@ module.exports = require("path");;
 /***/ ((module) => {
 
 "use strict";
-module.exports = require("stream");;
+module.exports = require("stream");
 
 /***/ }),
 
@@ -11781,7 +11798,7 @@ module.exports = require("stream");;
 /***/ ((module) => {
 
 "use strict";
-module.exports = require("tls");;
+module.exports = require("tls");
 
 /***/ }),
 
@@ -11789,7 +11806,7 @@ module.exports = require("tls");;
 /***/ ((module) => {
 
 "use strict";
-module.exports = require("url");;
+module.exports = require("url");
 
 /***/ }),
 
@@ -11797,7 +11814,7 @@ module.exports = require("url");;
 /***/ ((module) => {
 
 "use strict";
-module.exports = require("util");;
+module.exports = require("util");
 
 /***/ }),
 
@@ -11805,7 +11822,7 @@ module.exports = require("util");;
 /***/ ((module) => {
 
 "use strict";
-module.exports = require("zlib");;
+module.exports = require("zlib");
 
 /***/ })
 
@@ -11844,7 +11861,9 @@ module.exports = require("zlib");;
 /************************************************************************/
 /******/ 	/* webpack/runtime/compat */
 /******/ 	
-/******/ 	if (typeof __nccwpck_require__ !== 'undefined') __nccwpck_require__.ab = __dirname + "/";/************************************************************************/
+/******/ 	if (typeof __nccwpck_require__ !== 'undefined') __nccwpck_require__.ab = __dirname + "/";
+/******/ 	
+/************************************************************************/
 var __webpack_exports__ = {};
 // This entry need to be wrapped in an IIFE because it need to be in strict mode.
 (() => {
--- a/src/labeler.ts
+++ b/src/labeler.ts
@@ -241,6 +241,12 @@ function checkMatch(changedFiles: string[], matchConfig: MatchConfig): boolean {
     }
   }
 
+  if (matchConfig.branch !== undefined) {
+    if (!checkBranch(matchConfig.branch)) {
+      return false;
+    }
+  }
+
   return true;
 }
 
--- a/src/labeler.ts
+++ b/src/labeler.ts
@@ -214,6 +214,20 @@ function checkAll(changedFiles: string[], globs: string[]): boolean {
   return true;
 }
 
+function checkBranch(glob: string): boolean {
+  const matcher = new Minimatch(glob);
+  const branchName = github.context.ref;
+  core.debug(` checking "branch" pattern against ${branchName}`);
+  core.debug(`  - ${printPattern(matcher)}`);
+  if (!matcher.match(branchName)) {
+    core.debug(`   ${printPattern(matcher)} did not match`);
+    return false;
+  }
+
+  core.debug(`   "branch" pattern matched`);
+  return true;
+}
+
 function checkMatch(changedFiles: string[], matchConfig: MatchConfig): boolean {
   if (matchConfig.all !== undefined) {
     if (!checkAll(changedFiles, matchConfig.all)) {
--- a/src/labeler.ts
+++ b/src/labeler.ts
@@ -6,6 +6,7 @@ import { Minimatch, IMinimatch } from "minimatch";
 interface MatchConfig {
   all?: string[];
   any?: string[];
+  branch?: string;
 }
 
 type StringOrMatchConfig = string | MatchConfig;
--- a/__mocks__/@actions/github.ts
+++ b/__mocks__/@actions/github.ts
@@ -8,6 +8,7 @@ export const context = {
     owner: "monalisa",
     repo: "helloworld",
   },
+  ref: "test/testing-time",
 };
 
 const mockApi = {
--- a/__mocks__/@actions/github.ts
+++ b/__mocks__/@actions/github.ts
@@ -8,7 +8,6 @@ export const context = {
     owner: "monalisa",
     repo: "helloworld",
   },
-  ref: "test/testing-time",
 };
 
 const mockApi = {
--- a/__tests__/fixtures/branches.yml
+++ b/__tests__/fixtures/branches.yml
@@ -1,0 +1,5 @@
+test-branch:
+  - branch: "test/*"
+
+feature-branch:
+  - branch: "*/feature/*"
--- a/__tests__/main.test.ts
+++ b/__tests__/main.test.ts
@@ -105,6 +105,7 @@ describe("run", () => {
   });
 
   it("adds labels based on the branch names that match the glob pattern", async () => {
+    github.context.ref = "test/testing-time";
     usingLabelerConfigYaml("branches.yml");
     await run();
 
@@ -114,6 +115,20 @@ describe("run", () => {
       repo: "helloworld",
       issue_number: 123,
       labels: ["test-branch"],
+    });
+  });
+
+  it("adds labels based on branch names that match the glob pattern", async () => {
+    github.context.ref = "my/feature/that-i-like";
+    usingLabelerConfigYaml("branches.yml");
+    await run();
+
+    expect(addLabelsMock).toHaveBeenCalledTimes(1);
+    expect(addLabelsMock).toHaveBeenCalledWith({
+      owner: "monalisa",
+      repo: "helloworld",
+      issue_number: 123,
+      labels: ["feature-branch"],
     });
   });
 });
--- a/__tests__/main.test.ts
+++ b/__tests__/main.test.ts
@@ -15,6 +15,7 @@ const paginateMock = jest.spyOn(gh, "paginate");
 const getPullMock = jest.spyOn(gh.rest.pulls, "get");
 
 const yamlFixtures = {
+  "branches.yml": fs.readFileSync("__tests__/fixtures/branches.yml"),
   "only_pdfs.yml": fs.readFileSync("__tests__/fixtures/only_pdfs.yml"),
 };
 
@@ -101,6 +102,19 @@ describe("run", () => {
 
     expect(addLabelsMock).toHaveBeenCalledTimes(0);
     expect(removeLabelMock).toHaveBeenCalledTimes(0);
+  });
+
+  it("adds labels based on the branch names that match the glob pattern", async () => {
+    usingLabelerConfigYaml("branches.yml");
+    await run();
+
+    expect(addLabelsMock).toHaveBeenCalledTimes(1);
+    expect(addLabelsMock).toHaveBeenCalledWith({
+      owner: "monalisa",
+      repo: "helloworld",
+      issue_number: 123,
+      labels: ["test-branch"],
+    });
   });
 });
 
--- a/__tests__/main.test.ts
+++ b/__tests__/main.test.ts
@@ -118,7 +118,7 @@ describe("run", () => {
     });
   });
 
-  it("adds labels based on branch names that match the glob pattern", async () => {
+  it("adds labels based on branch names that match different glob patterns", async () => {
     github.context.ref = "my/feature/that-i-like";
     usingLabelerConfigYaml("branches.yml");
     await run();


amiel:
FYI: I found that this wasn't working for me as expected. I got it working for me in https://github.com/bentohq/release-drafter-spikes/pull/9. I didn't spend much time writing up any context, partly because this PR doesn't have any response, so I'm not sure if anyone is planning to take the time to review this. I'd be happy to communicate more about it if there's interest.

joshdales:
@amiel If it's not working properly I'm still interested so would appreciate any information you can give 😄

amiel:
Hi @joshdales. I just realized that I referenced the wrong pull request in my last comment. I've edited it to reference the one I meant to reference: https://github.com/joshdales/labeler/pull/1.

The problem I ran in to was that `github.context.ref` was producing something like `refs/pull/23/head`, which would not match my branch names like `feature/foo`.

Using `github.context.payload.pull_request.head.ref` resolves this for me and uses the branch name that I actually want to match on. I don't have a lot of experience with github actions, but I found this object path here: https://github.com/TimonVS/pr-labeler-action/blob/master/src/action.ts#L25.

amiel:
I also considered updating your changes to support a list of branch patterns, just like the way lists of files are supported. However, I got it working by using the glob matching and I added a test to demonstrate that. This took me a little while to figure out because the minimatch documentation shows an example with `(bar|foo)`, but I needed to use the glob syntax (`{bar,foo}`) instead.

joshdales:
@amiel took your suggestion and I've updated this PR to support a list of branch arguments as well 👍

amiel:
@joshdales Awesome! I'll pull those changes upstream too since we are pinned to our branch until this gets actually released.

davidh-unmind:
Be cool to get this feature shipped. I'd selfishly find it useful.

dist/index.js
```diff
@@ -92,6 +92,14 @@ function getPrNumber() {
     }
     return pullRequest.number;
 }
+function getBranchName() {
```
DennisBoanini, line 95: I think we can get the branch name using `github.head_ref`

https://docs.github.com/en/actions/learn-github-actions/contexts#github-context

DennisBoanini (approved):

Alvaro-Campesino:
You should update the README.md file to documenting your feature.

justin-kidman-axomic:
Any updates on this?

lucacome:
Any chance we can get this merged soon?

arivera-xealth:
Any updates @DennisBoanini @pje ?

DennisBoanini:
> Any updates @DennisBoanini @pje ?

I don't know, the owner of the PR is @joshdales, I only add a suggestion

DennisBoanini:
I see that there is 1 pending reviewer and the branch have conflicts. Any change to merge this PR?

joshdales:
@DennisBoanini @arivera-xealth I tried to keep it updated and added to it for a while but it felt kinda fruitless after 5 months 😞 

Happy to come back and make the suggestions and fix those conflicts once a code owner has looked at it

franklinkim:
@joshdales review received. would love to see this being merged.

joshdales:
@franklinkim the review isn't from a code owner so it can't be merged 😢

franklinkim:
😭

DennisBoanini:
Guys, why this PR still cannot be merged?

joshdales:
@DennisBoanini I can't merge it without it being reviewed by a code owner first.

I've fixed the merge conflict though and added some notes to the readme to try and breathe some life back into this PR. 

@pje Any chance that you or someone else on the actions team could take a look?

adilnaimi:
@pje 👇 
Any chance that you or someone else on the actions team could take a look?

DennisBoanini:
> I wonder what's going on with this repo and the maintainers. This PR was opened more than a year ago and there's clearly a lot of interest in the feature, but not even one response from the maintainers. 😞

You're right, that's incredible. I don't understand why.

DennisBoanini:
Who is the one that can accept this PR?

elendil-software:
> I wonder what's going on with this repo and the maintainers. This PR was opened more than a year ago and there's clearly a lot of interest in the feature, but not even one response from the maintainers. 😞

Also interested in this feature

maistrotoad:
@pje any chance you want to take this up in your project? It looks like a really nice addition I would definitely use

lucacome:
maybe @chrispat ?

amiel:
Hi @joshdales. Thanks for maintaining this branch despite the lack of review from an actual code owner. I know it has been frustrating. I appreciate it as my team and I are still using this in daily builds.

Something happened with the most recent commit ([0ad789c](https://github.com/actions/labeler/pull/203/commits/0ad789cf22fa7557b0bc464ae9424fcfd6b0ffa3)) and we're getting this error:

```
| /run/act/actions/joshdales-labeler@main/dist/index.js:14965
| util.inherits(TunnelingAgent, events.EventEmitter);
|      ^
| 
| TypeError: util.inherits is not a function
```

Also, it looks like my changes from joshdales#1 are not included anymore. I have hard-coded our action to use joshdales/labeler@6e276063fecb6573bf5dc89df1f4a1e317e3a6ce, which is still working for me.

Anyway, I just wanted to let you know that something doesn't seem right with that last merge. Thanks again.

joshdales:
@amiel Ahh damn - that'll teach me to merge late at night 😞. 

Thanks for letting me know I'll take a look at this tonight and try and get it fixed 👍

amiel:
Thanks @joshdales!

joshdales:
Okay I've got it working again! 🎉 https://github.com/joshdales/glowing-octo-disco/pull/1

Seems that I just needed to run the build command again

amiel:
That did the trick. Thanks so much!

Beshr-wittra (approved):

amiel (approved):

DennisBoanini:
Who is @actions/actions-runtime that is requested for the approval?

joshdales:
> Who is @actions/actions-runtime that is requested for the approval?
@DennisBoanini 

That's the team that are code owners of this repo, so in theory any of them should be able to review this. 

From what I can tell from snooping other PRs over the past year that this one has been open @pje seems to be the one that mostly merges them, though maybe @cory-miller or @thboop could take a look as it looks like they are also code owners 😅

README.md
```diff
@@ -82,6 +88,10 @@ source:
 frontend:
 - any: ['src/**/*.js']
   all: ['!src/main.js']
+  
+ # Add 'feature' label to any branch that starts with `feature` or has a `feature` section in the name
+ feature:
+ - branch: ['feature/**', '*/feaure/**']
```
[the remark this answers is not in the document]
joshdales: 😆 Yeah it should, thanks for catching that

jmmarco:
We would appreciate if somebody from the team could look into this @pje 🙂

joshdales:
@MaksimZhukov Are you the new maintainer of this project? Would it be possible to get a review on this PR please?

panticmilos:
hi @joshdales, i am maintaining this repo as well. We will take a look into this PR

MaksimZhukov:
Hello @joshdales!
Sorry for the late response and thank you for your patience and contribution!
We will review this PR in the near future.

src/labeler.ts
```diff
@@ -6,6 +6,7 @@ import {Minimatch, IMinimatch} from 'minimatch';
 interface MatchConfig {
   all?: string[];
   any?: string[];
+  branch?: string | string[];
```
MaksimZhukov, line 9: At the moment, the new field name in the yaml config file (`branch`) is not specific. We should make it more concrete so that there is no confusion and users can understand what type of branch is meant. How about `headBranch`? 
There is a [feature request](https://github.com/actions/labeler/issues/54) in the repository to add similar functionality, but for base branches. It would be great if you could add this in scope of your PR. I think we can reuse some logic that was introduced by you.

MaksimZhukov, line 9: At the moment, the new field (`branch`) is on the same level as the patterns for changed files. The current structure of the configuration file looks like this:
```
- label name:
  - patterns for changed files
  - branch:
    - patterns for branch
```

Potentially, more new fields will be added in the future and such a structure will become non-obvious, non-intuitive. I think we should introduce a new field for changed files as well:
```
- label name:
  - changed files:
    - patterns
  - branch:
    - patterns
... 
```
This will be a breaking change, but we're going to release a new major version anyway to fix the [bug](https://github.com/actions/labeler/pull/480) related to the `sync-labels` input. We can include both changes in the new major release.

src/labeler.ts
```diff
@@ -213,6 +223,42 @@ function checkAll(changedFiles: string[], globs: string[]): boolean {
...
+function checkBranch(glob: string | string[]): boolean {
+  const branchName = getBranchName();
+  if (!branchName) {
+    core.debug(` no branch name`);
+    return false;
+  }
+
+  core.debug(` checking "branch" pattern against ${branchName}`);
+  if (Array.isArray(glob)) {
+    const matchers = glob.map(g => new Minimatch(g));
```
MaksimZhukov, line 246: You are using the minimatch library under the hood, and it works fine, but as far as I know, it is designed mainly to work with paths, and its syntax may not be obvious to some users for branch filtering. I think the best option here would be to just use regular expressions. We have a [feature request](https://github.com/actions/labeler/issues/55) to add labels based on the title and description of the pull request and we could prepare a foundation for this functionality in scope of your PR.

MaksimZhukov (commented):
Hello @joshdales!
I have reviewed your PR, thank you for it! I left a few comments, please take a look.
I haven't looked at the tests and documentation, I think we should agree on the approach first.

src/labeler.ts
joshdales: Interesting so to follow up from your previous comment the we would actually wanna change this to the following:
```yml
- label name:
  - changed-files:
    - patterns
  - head-branch:
    - patterns 
```
So that we could add another level for `base-branch` in the future? 

🤔 Or in the way that `changed-files` can have `any` and `all` we keep that level called `branch` as you suggest here and then add `head` and `base` options to that. So the config might end up looking like this:
```yml
- label name:
  - changed-files:
    - any: 
      - patterns
    - all:
      - patterns
  - branch:
    - head:
      - patterns
    - base:
      - patterns 
```
I like that idea 👍

joshdales: Small follow up question to that - what would we want the defaults to be in the case that these nested options are not provided? 
eg.
```yml
- label 1: 'pattern'
- label 2:
  - changed-files: 'pattern'
- label 3:
  - branch: 'pattern'
```

Currently it defaults to match `any` file that has changed, would we still want to keep that as it is or adjust it? Based on how things are currently I guess it might be something like this:
- label-1 matches any file that has changed?
- label-2 also matches any file that has changed?
- label-3 matches the head branch?

MaksimZhukov: In my understanding, the final structure should look like this:
```yml
- label name:
  - changed-files:
    - patterns
    # or
    - any: 
      - patterns
    - all:
      - patterns
  - head-branch:
    - patterns
  - base-branch:
    - patterns 
```
We may want to expand branch settings, for example the base and/or head branches can have `include` or `exclude` options, so I expect them to be separate fields so that we can avoid deep nesting:
```yml
...
  - head-branch:
    - include:
      - patterns
    - exclude:
      - patterns
  - base-branch:
    - include:
      - patterns
    - exclude:
      - patterns
...
```
Regarding your question about omitting nested options - I think we should adhere the following approach:
- `label 1` - throw an error, because it is not obvious what these patterns are related to, what exactly the user meant.
- `label 2` - matches any file that has changed, as you mentioned
- `label 3` - Separating the branches, as I suggested, would simplify the logic:
```yml
  - head-branch:
    - patterns
  - base-branch:
    - patterns 
```

joshdales: Would having include and exclude options more not come from the pattern that you pass? Regardless avoiding deeply nested objects makes sense for now so I'll start on adding that functionality 👍

joshdales:
@amiel and anyone else that is using this using this branch in your action - I am currently working on updating it based on PR feedback which restructures the yml config file and it's going to break this version, so you probably wanna pin your action to my last commit: `7b1327b4c44a8794dfc7573d60637cd60ce4b697` to keep using this as is until it's ready for release.

I'm doing the work on another branch and I have a draft PR open for it https://github.com/joshdales/labeler/pull/3, so it's not going to break instantly but I'll give you another warning when I going to merge it.

MaksimZhukov:
Hello @joshdales!
Could you pease clarify the current status of this PR? When do you plan to prepare the necessary changes?
Do you need help / do you want to discuss something else?

joshdales:
@MaksimZhukov I have updated this PR!

Changes include:
- Changing the match config structure so there are now 3 main options: `changed-files`, `base-branch`, `head-branch`. `any` and `all` options are now nested under `changed-files`.
- A new `branch.ts` file that handles which is where all the branch logic lives. Along with adding unit tests for this as well.
- Renaming a few functions in `labeler.ts` to differentiate between the changing files and branches.
- I reversed the conditions of `checkMatch` to make the changes from the matching of files a bit easier without having to refactor more of the logic around the parsing of the yaml config file.
- Removed the deprecated `IMinimatch` interface

I tested it out with this PR: https://github.com/joshdales/glowing-octo-disco/pull/4 and it is working. Hope all these changes make sense to you 🙏

jeniabrook:
@joshdales super excited for this to land. 

I'm already using labeler almost in every repo/monorepo that I have, this would enhance the DX even more. 

Thanks!

README.md
```diff
@@ -26,13 +26,15 @@ For more control over matching, you can provide a match object instead of a simp
 ```yml
 - any: ['list', 'of', 'globs']
   all: ['list', 'of', 'globs']
+  branch: ['list', 'of, 'globs']
```
jvanderwee, line 29: should this and below be `base-branch` / `head-branch`?

joshdales: I haven't pushed my updates to the readme yet

joshdales: it's pushed now 😄

src/labeler.ts
```diff
@@ -122,10 +127,10 @@ async function fetchContent(
...
-function getLabelGlobMapFromObject(
+function getLabelConfigMapFromObject(
   configObject: any
-): Map<string, StringOrMatchConfig[]> {
-  const labelGlobs: Map<string, StringOrMatchConfig[]> = new Map();
+): Map<string, MatchConfig[]> {
+  const labelGlobs: Map<string, MatchConfig[]> = new Map();
   for (const label in configObject) {
     if (typeof configObject[label] === 'string') {
       labelGlobs.set(label, [configObject[label]]);
```
MaksimZhukov, line 136: I think we can get rid of this condition since a simple string (`label: 'pattern'`) is no longer a valid glob. We should throw the error in this case.

src/labeler.ts
```diff
@@ -33,16 +38,16 @@ export async function run() {
...
       configPath
     );
 
     const labels: string[] = [];
     const labelsToRemove: string[] = [];
-    for (const [label, globs] of labelGlobs.entries()) {
+    for (const [label, configs] of labelConfigs.entries()) {
       core.debug(`processing ${label}`);
-      if (checkGlobs(changedFiles, globs)) {
+      if (checkGlobs(changedFiles, configs)) {
```
MaksimZhukov, line 50: You are passing `configs`, but the function still accepts `globs`. We should stick to consistency.

src/labeler.ts
```diff
@@ -141,23 +146,65 @@ function getLabelGlobMapFromObject(
...
+}
+
+function printPattern(matcher: Minimatch): string {
   return (matcher.negate ? '!' : '') + matcher.pattern;
 }
 
 export function checkGlobs(
   changedFiles: string[],
-  globs: StringOrMatchConfig[]
+  globs: MatchConfig[]
```
MaksimZhukov, line 207: Currently, we can only specify one config for a label (`changed-files` or `base-branch` or `head-branch`). It would be great if we could provide more flexible behavior so that users can combine configs. The following update should solve it:
```ts
export function checkGlobs(
  changedFiles: string[],
  globs: MatchConfig[]
): boolean {
  for (const glob of globs) {
    core.debug(` checking pattern ${JSON.stringify(glob)}`);
    const matchConfig = toMatchConfig(glob);
    if (!checkMatch(changedFiles, matchConfig)) {
      return false;
    }
  }
  return true;
}
```

src/labeler.ts
```diff
@@ -214,19 +261,31 @@ function checkAll(changedFiles: string[], globs: string[]): boolean {
...
+      return true;
     }
   }
 
-  if (matchConfig.any !== undefined) {
-    if (!checkAny(changedFiles, matchConfig.any)) {
-      return false;
+  if (matchConfig.changedFiles?.any !== undefined) {
+    if (checkAny(changedFiles, matchConfig.changedFiles.any)) {
+      return true;
```
MaksimZhukov, line 272: This change introduces a regression. There is an example in the README file:
```yml
# Add 'frontend` label to any change to *.js files as long as the `main.js` hasn't changed
frontend:
- any: ['src/**/*.js']
  all: ['!src/main.js']
```

Now, if `all` passes, `any` will not be checked.

src/branch.ts
```diff
@@ -0,0 +1,83 @@
...
+    const patterns = branchConfig.baseBranch;
+    if (typeof patterns === 'string') {
+      branchConfig.baseBranch = [patterns];
+    }
+  }
+
+  return branchConfig;
+}
+
+export function getBranchName(branchBase?: BranchBase): string | undefined {
```
MaksimZhukov, line 38: Could you please clarify why the parameter is optional?

__tests__/branch.test.ts
```diff
@@ -0,0 +1,155 @@
...
+        expect(result).toMatchObject<BranchMatchConfig>({
+          headBranch: ['testing']
+        });
+      });
+    });
+  });
+
+  describe('when the config contains a base-branch option', () => {
+    const config = {'base-branch': ['testing']};
+    it('sets headBranch in the matchConfig', () => {
```
MaksimZhukov, line 126:
```suggestion
    it('sets baseBranch in the matchConfig', () => {
```

__tests__/branch.test.ts
```diff
@@ -0,0 +1,155 @@
...
+      const result = toBranchMatchConfig(config);
+      expect(result).toMatchObject<BranchMatchConfig>({
+        baseBranch: ['testing']
+      });
+    });
+
+    describe('and the matching option is a string', () => {
+      const stringConfig = {'base-branch': 'testing'};
+
+      it('sets headBranch in the matchConfig', () => {
```
MaksimZhukov, line 136:
```suggestion
      it('sets baseBranch in the matchConfig', () => {
```

__tests__/branch.test.ts
```diff
@@ -0,0 +1,155 @@
...
+        expect(result).toMatchObject<BranchMatchConfig>({
+          baseBranch: ['testing']
+        });
+      });
+    });
+  });
+
+  describe('when the config contains both a base-branch and head-branch option', () => {
+    const config = {'base-branch': ['testing'], 'head-branch': ['testing']};
+    it('sets headBranch in the matchConfig', () => {
```
MaksimZhukov, line 147:
```suggestion
    it('sets headBranch and baseBranch in the matchConfig', () => {
```

README.md
```diff
@@ -9,40 +9,48 @@
...
-Create a `.github/labeler.yml` file with a list of labels and [minimatch](https://github.com/isaacs/minimatch) globs to match to apply the label.
+Create a `.github/labeler.yml` file with a list of labels and config options to match and apply the label.
 
-The key is the name of the label in your repository that you want to add (eg: "merge conflict", "needs-updating") and the value is the path (glob) of the changed files (eg: `src/**/*`, `tests/*.spec.js`) or a match object.
+The key is the name of the label in your repository that you want to add (eg: "merge conflict", "needs-updating") and the value is a match object.
 
 #### Match Object
 
-For more control over matching, you can provide a match object instead of a simple path glob. The match object is defined as:
+The match object allows control over the matching options, you can specify the label to be applied based on the files that have changed or the name of the branch. For the changed files options you provide a path glob, and a regexp for the branch names.
```
MaksimZhukov, line 24: I think it's still necessary to point users to the minimatch documentation
```suggestion
The match object allows control over the matching options, you can specify the label to be applied based on the files that have changed or the name of the branch. For the changed files options you provide a [path glob](https://github.com/isaacs/minimatch#minimatch), and a regexp for the branch names.
```

MaksimZhukov:
Could you please also update [this example](https://github.com/actions/labeler#create-workflow) with the new major version of action:
```yml
...
- uses: actions/labeler@v5
...
```

src/branch.ts
joshdales: I guess that was my own personal bias for the use case that I originally created this PR for and the original implementation - but yeah it doesn't really need to be optional, I'll remove that and just make it always accept an argument

joshdales: I guess that was my own personal bias for the use case that I originally created this PR for and the original implementation - but yeah it doesn't really need to be optional, I'll remove that and just make it always accept an argument

src/labeler.ts
```diff
@@ -214,19 +261,31 @@ function checkAll(changedFiles: string[], globs: string[]): boolean {
...
     }
   }
 
-  return true;
+  if (matchConfig.headBranch !== undefined) {
+    if (checkBranch(matchConfig.headBranch, 'head')) {
+      return true;
+    }
+  }
```
dmitry-shibanov, line 280: Could you please change it to:
```ts
  if (matchConfig.headBranch) {
    return checkBranch(matchConfig.headBranch, 'head');
  }
```
and the same for others.

src/labeler.ts
```diff
@@ -141,23 +146,65 @@ function getLabelGlobMapFromObject(
...
+      // so assign them to our `changedFilesMatchConfig`
+      changedFilesConfig.forEach(config => {
+        // Make sure that the values that we assign to our match config are an array
+        Object.entries(config).forEach(([key, value]) => {
+          const element = {
+            [key]: Array.isArray(value) ? value : [value]
+          };
+          Object.assign(changedFilesMatchConfig.changedFiles, element);
+        });
+      });
```
dmitry-shibanov, line 184: I think we can change it to these lines:
```ts
      Object.assign(changedFilesMatchConfig.changedFiles, ...changedFilesConfig);
      Object.keys(changedFilesMatchConfig.changedFiles).forEach(key => {
        const value = changedFilesMatchConfig.changedFiles[key];
        changedFilesMatchConfig.changedFiles[key] = Array.isArray(value) ? value : [value]
      });
```

src/labeler.ts
joshdales: I was just following the existing pattern that was here, but I'm more than happy to update that as I prefer that syntax personally 👍

macksal: Does this break if both a head branch and a base branch are present? In that case, I think we would need to do both checks

README.md
```diff
@@ -9,40 +9,48 @@
...
+The match object is defined as:
 
 ```yml
-- any: ['list', 'of', 'globs']
-  all: ['list', 'of', 'globs']
+- changed-files:
+  - any: ['list', 'of', 'globs']
+  - all: ['list', 'of', 'globs']
+- base-branch: ['list', 'of, 'regexps']
+- head-branch: ['list', 'of, 'regexps']
```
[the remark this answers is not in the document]
IvanZosimov: Just a lack of single quotes in the example.

README.md
```diff
@@ -52,41 +60,60 @@ From a boolean logic perspective, top-level match objects are `OR`-ed together a
...
-- '**/*.txt'
+- changed-files: '**/*.txt'
 
+# Add 'label4' to any PR where the head branch name starts with 'example4'
+label4:
+- head-branch: '^example4/**'
+
+# Add 'label5' to any PR where the base branch name starts with 'example5'
+label5:
+- base-branch: '^example4/'
```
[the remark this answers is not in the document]
IvanZosimov: I believe slash is unneeded here.

src/branch.ts
```diff
@@ -0,0 +1,83 @@
...
+      branchConfig.headBranch = [patterns];
+    }
+  }
+
+  if (branchConfig.baseBranch) {
+    const patterns = branchConfig.baseBranch;
+    if (typeof patterns === 'string') {
+      branchConfig.baseBranch = [patterns];
+    }
+  }
```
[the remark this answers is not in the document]
IvanZosimov: I believe we can simplify these lines of code.

MaksimZhukov:
Hi @joshdales!
What is the current status of your progress? Do you need any help?

joshdales:
Hi @MaksimZhukov working through all the comments at the moment.

I'm wondering if it might be better to extract out all the changed files checking logic into its own file similar to what I did with the branch stuff. I resisted it initially because it would have made the diff huge, but as it already looks like I'm going to be touching a lot might be best to do it now, if you are okay with that?

MaksimZhukov:
Thank you @joshdales !
This is a great idea, but if it takes a lot of time / effort, then you can leave it as it is and we will prepare these changes in a separate PR.

joshdales:
@MaksimZhukov @dmitry-shibanov @IvanZosimov I've updated the PR and hopefully addressed all your comments, thanks for all the suggestions. 👍

I've moved all the logic todo with labelling based on changing files into it's own file (`changedFiles.ts`) and added some of unit tests for that file and the main functions that get used from it in `labeller.ts`. 

I didn't realise that it only let's me re-request one of you at a time so sorry about spamming you with re-request notifications 🤦

I haven't tested my changes yet with an external repo - I'll try and get to that shortly.

dmitry-shibanov (approved):

maurolemos:
@joshdales almost 2years... congrats for the persistence! 🎉

MaksimZhukov (commented):
@joshdales Thanks for the update! It looks great! 
I left some small comments, please take a look

src/labeler.ts
```diff
@@ -122,106 +105,72 @@ async function fetchContent(
...
 
-// equivalent to "Array.some()" but expanded for debugging and clarity
-function checkAny(changedFiles: string[], globs: string[]): boolean {
-  const matchers = globs.map(g => new Minimatch(g));
-  core.debug(`  checking "any" patterns`);
-  for (const changedFile of changedFiles) {
-    if (isMatch(changedFile, matchers)) {
-      core.debug(`  "any" patterns matched against ${changedFile}`);
-      return true;
+function checkMatch(changedFiles: string[], matchConfig: MatchConfig): boolean {
```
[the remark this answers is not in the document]
joshdales: I actually had changed the conditions at one point, but [you pointed out that it introduced a regression](https://github.com/actions/labeler/pull/203#discussion_r1124428162) so I reverted that. 

🤔 Maybe a guard clause that checks if the match config is empty and returns false before doing any checks though I might have to adjust `toChangedFilesMatchConfig` to make sure that it doesn't add any extra keys that we don't currently allow for.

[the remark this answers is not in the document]
joshdales: Okay - I have added some more tests around this kind of thing and made sure that we don't set extra items in the changed files match config, so even if there are extra properties in the yaml file it shouldn't accidentally apply the label.

dmitry-shibanov:
Hello @joshdales. Could you please regenerate dist with the `npm run build` command.

src/labeler.ts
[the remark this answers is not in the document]
MaksimZhukov: Yes, you're right, I completely forgot. 
> I actually had changed the conditions at one point, but https://github.com/actions/labeler/pull/203#discussion_r1124428162 so I reverted that.

You have found a good solution!

IvanZosimov (approved):

Ameausoone:
So close ! 😅 

Great job BTW ! 🚀

MaksimZhukov:
Thank you @joshdales ! We will merge the PR and release a new action version on Monday

lucacome:
I've been using the action from this PR for a while (before the recent changes) and I had a config like this:
```
tests:
- branch: ['tests/**', 'test/**']
- tests/**/*
- perf-tests/**/*

dependencies:
- branch: ['deps/**', 'dep/**', 'dependabot/**']
- go.mod
- go.sum
```

now I've changed it to this to test it before the release on Monday:
```
tests:
  - head-branch: ['^tests/', '^test/']
  - changed-files:
    - any: ['tests/**/*', 'perf-tests/**/*']

dependencies:
  - head-branch: ['^deps/', '^dep/', '^dependabot/']
  - changed-files:
    - any: ['go.mod', 'go.sum']
```

and let's say have I have branch called `deps/some-update` that also updates files in the folder `tests/`, before I would get both the label `tests` and `dependencies`.
With the new config I don't get any labels...am I missing something? Did I not understand how this is supposed to work?

These are the logs:
```
...
##[debug]processing tests
##[debug] checking config {"headBranch":["^tests/","^test/"]}
##[debug] checking "branch" pattern against deps/fff
##[debug]  - /^tests\//
##[debug]   /^tests\// did not match
##[debug]  - /^test\//
##[debug]   /^test\// did not match
##[debug]  "branch" patterns did not match against deps/fff
...
##[debug]processing dependencies
##[debug] checking config {"headBranch":["^deps/","^dep/","^dependabot/"]}
##[debug] checking "branch" pattern against deps/fff
##[debug]  - /^deps\//
##[debug]   "branch" pattern matched
##[debug]  "branch" patterns matched against deps/fff
##[debug] checking config {"changedFiles":{"any":["go.mod","go.sum"]}}
##[debug]  checking "any" patterns
##[debug]    matching patterns against file tests/test
##[debug]   - go.mod
##[debug]   go.mod did not match
##[debug]  "any" patterns did not match any files
```
it seems like it's not even processing the files for `tests`

joshdales:
@lucacome what is happening is that when you combine different keys they all have to match 
which is the current behaviour according to the Readme: https://github.com/actions/labeler/blob/main/README.md?plain=1#L24
> Unlike the top-level list, the list of path globs provided to any and all must ALL match against a path for the label to be applied.

So the branch and changed files all need to match, and the files don't get processed in the `tests` example is because the branch check just failed.

If you wanted to get this to work now I think that you would just need to split up the keys:
```yml
tests:
  - head-branch: ['^tests/', '^test/']

tests:
  - changed-files:
    - any: ['tests/**/*', 'perf-tests/**/*']

dependencies:
  - head-branch: ['^deps/', '^dep/', '^dependabot/']

dependencies:
  - changed-files:
    - any: ['go.mod', 'go.sum']
```

🤔 I think that a fact that it used to work was an error with my previous implementation.

lucacome:
@joshdales I thought about duplicating the labels, but it makes the YAML invalid
```
Error: YAMLException: duplicated mapping key (51:1)
```

The line you quoted is a little bit confusing to me, not really sure what the "top-level list" is 😅


I assumed something like this would be a list of different conditions for the label to be applied
```
dependencies:
  - head-branch: ['^deps/', '^dep/', '^dependabot/']
  - changed-files:
    - any: ['go.mod', 'go.sum']
 ```
and not that all of them need to be true at the same time, IMHO it's not as useful...

**Comment by joshdales** [comment]
Ahhhh good point about the yaml. Damn.

My interpretation of the "top level" list is that it has multiple keys so in that example `any` and `all` or in this case `head-branch` and `changed-files` as those are the highest level keys now. I actually didn't realise that this was the way it was supposed to work until the review and I introduced a regression 😆.

I would be inclined to agree with you that that is how I would expect it work. Probably a call for @MaksimZhukov and the maintainers of this action. 🤔 though if this is going to be a breaking version anyway maybe a more obvious use case is better than flexibility of combining rules?

**Comment by joshdales** [comment]
🤔 Actually maybe the config is defined in the wrong way - maybe `any` and `all` should be the top level keys and the branches and changed file options should be nested.

```yml
label:
  - any: 
    - changed-files: ['list', 'of', 'globs']
    - base-branch: ['list', 'of', 'regexps']
    - head-branch: ['list', 'of', 'regexps']
  - all:
    - changed-files: ['list', 'of', 'globs']
    - base-branch: ['list', 'of', 'regexps']
    - head-branch: ['list', 'of', 'regexps']
```

Defaulting to the `any` option if neither `all` or `any` are provided.

I think if you did that if would give you the flexibility to mix rules of the current implementation but also not be as confusing.

joshdales:
@lucacome Okay, I think that I've got this new structure working, opened a PR against this branch here: https://github.com/joshdales/labeler/pull/4, all the tests are passing but there is still some clean up that I need to do, which I will try and get to this weekend.

Please note that I haven't tested it on a real PR yet 😅

@MaksimZhukov maybe you should hold off merging this PR as it currently is on Monday?

lucacome:
@joshdales I've tested the changes in that PR and everything works like I originally thought it would 🚀

MaksimZhukov:
@lucacome thank you for the feedback!

@joshdales we will postpone merging this PR and review the new one, thanks for the updates!

joshdales:
@MaksimZhukov I updated this PR with the new changed proposed and new config structure, added some more tests as well - so hopefully it will all makes sense 👍

MaksimZhukov:
Thank you @joshdales ! We will take a look

lucacome:
Did we lose the momentum here? We were so close... 😭

MaksimZhukov:
Hello @joshdales ! Sorry for the late response and thanks for your patience!
I would like to clarify one thing. As described in the documentation, the `any` field works as follows: ALL globs must match against ANY changed path. You changed this behaviour with the latest update, now ANY globs must match against ANY changed path. Did you do this intentionally? If so, could you please explain why? Do you think this is a useless use case?

MaksimZhukov:
Thank you @joshdales for the explanation! 

I agree with you that the `any` selector can confuse users, and it is, but I believe that this problem is due to poor documentation. We will definitely rewrite documentation to make it more clear.

I think it would be great if we could leave the existing logic as it is, because customers use the `any` selector in a certain way and rely on the current behaviour. If we change the logic, we will deprive users of the current functionality, they will no longer be able to use it. Meanwhile, they can achieve the behaviour that you implement using simple strings
```yml
LabelName:
- any: 
  - changed-files:
    - pattern
    - pattern
```
Does it make sense to you?

joshdales:
In my head this wasn't really depriving users of existing functionality because it was making it easier to get an `any` match, plus it was a breaking change.

Regardless, I get what you are saying and I'll adjust the `changed-files` matching logic back to how it was previously 👍

lucacome:
Sorry everybody for the noise, but I'm back 😬

I was trying to adjust my config to the latest changes but it seems like I can't make something like this work anymore after the last 5 commits
```
tests:
  - any:
    - head-branch: ['^tests/', '^test/']
    - changed-files: ['tests/**/*', 'perf-tests/**/*']
  - all:
    - changed-files: ['!tests/requirements.txt', '!perf-tests/requirements.txt']
 ```

Is there still a way to achieve this that I can't see? I have a file like `tests/foo` that was changed, but the label is not applied, probably because of the old way `any` works that I never understood tbh.

If this is not possible anymore I'd like for it to be reverted to what @joshdales had before, since as he said this will be a breaking change version anyway.

But more than anything I hope this PR won't be here to see its two-year anniversary 🤣

**Comment by joshdales** [comment]
I think that in order to get the same behaviour you want your config to be like this:
```yml
tests:
  - any:
      - head-branch: ['^tests/', '^test/']
      - changed-files: 'tests/**/*'
      - changed-files: 'perf-tests/**/*'
  - all:
      - changed-files: ['!tests/requirements.txt', '!perf-tests/requirements.txt']
```
Though this is pretty obtuse and not obvious as to how to use it. To me it feels like this issue about the `any` changed file option is probably going to keep coming up and it looks there are currently a couple of other open issues that are related to this (#545, #101) which are already labeled as `bug` 🤷

MaksimZhukov:
Hi @joshdales and @lucacome! Sorry for the delay in responding!

We had a discussion within the team and we decided to follow your recommendations. Let's update the `any` selector to match any file against any glob pattern. We will release these changes as a new beta version so that users can test them and share their feedback. If they are satisfied with the new approach, we will release a stable version.

@joshdales please revert your recent changes. Sorry for going back and forth, and thank you for your time and efforts!

src/labeler.ts
```diff
@@ -122,110 +115,178 @@ async function fetchContent(
...
-    }
+export function checkAll(
+  matchConfigs: BaseMatchConfig[],
+  changedFiles: string[]
+): boolean {
+  core.debug(`  checking "all" patterns`);
+  if (!Object.keys(matchConfigs).length) {
+    core.debug(`  no "all" patterns to check`);
+    return false;
   }
```
MaksimZhukov, line 267: I guess this validation never works the way it should since `matchConfigs` is an array. The length of this array `[{}]` will be `1` and the `checkAll` function will return `true`

src/branch.ts
```diff
@@ -0,0 +1,98 @@
...
+  return false;
+}
+
+export function checkAllBranch(
+  regexps: string[],
+  branchBase: BranchBase
+): boolean {
+  const branchName = getBranchName(branchBase);
+  if (!branchName) {
+    core.debug(`   no branch name`);
```
MaksimZhukov, line 72: I think this message will be more helpful:
```suggestion
    core.debug(`   cannot fetch branch name from the pull request`);
```

__tests__/changedFiles.test.ts
```diff
@@ -0,0 +1,106 @@
...
+  toChangedFilesMatchConfig
+} from '../src/changedFiles';
+
+jest.mock('@actions/core');
+jest.mock('@actions/github');
+
+describe('checkAllChangedFiles', () => {
+  const changedFiles = ['foo.txt', 'bar.txt'];
+
+  describe('when the globs match every file that has changed', () => {
```
MaksimZhukov, line 14:
```suggestion
  describe('when the globs match every file that has been changed', () => {
```

MaksimZhukov (commented):
Thank you @joshgoebel for the updates! Everything looks great, I left some minor comments

IvanZosimov (approved):

src/labeler.ts
MaksimZhukov: Sorry for the vague description of the problem! I meant that the `toMatchConfig` function always returns an object so `matchConfigs` cannot be empty. The returned object can be empty if we specify invalid config, for example this
```yml
LabelName:
- all:
  - invalid-option: ['**']
```
will lead to `{"all":[{}]}` and length will be `1`.

The following update should fix the problem:
```js
  const ALLOWED_CONFIG_KEYS = ['changedFiles', 'headBranch', 'baseBranch'];

  if (!matchConfigs.some(configOption => ALLOWED_CONFIG_KEYS.includes(Object.keys(configOption)[0]))) {
    core.debug(`  no "all" patterns to check`);
    return false;
  }
```

joshdales: Ahh I see, thanks! I've updated this now 👍

lucacome:
I really didn't want to add another comment, but...seems like something is not working... 😭 

with the same config as usual:
```
tests:
  - any:
      - head-branch: ['^tests/', '^test/']
      - changed-files: ['tests/**/*', 'perf-tests/**/*']
  - all:
      - changed-files: ['!tests/requirements.txt', '!perf-tests/requirements.txt']
```

I get 
```
##[debug]processing tests
##[debug] checking config {"any":[{"headBranch":["^tests/","^test/"]},{"changedFiles":["tests/**/*","perf-tests/**/*"]}]}
##[debug]  checking "any" patterns
##[debug]   checking "branch" pattern against deps/foo
##[debug]    - /^tests\//
##[debug]    /^tests\// did not match
##[debug]    - /^test\//
##[debug]    /^test\// did not match
##[debug]   "branch" patterns did not match against deps/foo
##[debug]    matching patterns against file README.md
##[debug]     - tests/**/*
##[debug]     - perf-tests/**/*
##[debug]    no patterns matched
##[debug]    matching patterns against file tests/bar
##[debug]     - tests/**/*
##[debug]    tests/**/* matched
##[debug]   "any" patterns matched against tests/bar
##[debug] checking config {"all":[{"changedFiles":["!tests/requirements.txt","!perf-tests/requirements.txt"]}]}
##[debug]  checking "all" patterns
##[debug]  no "all" patterns to check
```

and the `tests` label is removed even tho `"any" patterns matched against tests/bar`.
It seems to me that is something with the "all" pattern from `no "all" patterns to check`.

marko-zivic-93 (approved):

MaksimZhukov:
Hello everyone!
We released the new version of the action that includes these changes. For more information, see the [release notes](https://github.com/actions/labeler/releases/tag/v5.0.0-alpha.1). 

@joshdales thanks for working on this PR, for your time and contribution! It was a pleasure to work with you! 
Thank you all for your comments and patience!

PR #314, opened 2022-02-02
Clarify `permissions` required to use Action

jamietanna:
As a means to reduce the privileges provided to the job executing, we
can document what the minimal `permissions` required for the Actions'
`GITHUB_TOKEN` is.


pje:
Thanks for the contribution, @jamietanna!

PR #350, opened 2022-04-20
Provide GH Token by default

dapirian:
With this change, users no longer need to explicitly pass a `repo-token`


dapirian:
@pje how does one find a code reviewer for these?

MichaelDeBoey:
Duplicate of #227

Issue #123, opened 2021-02-02
Issue when share workflow within organization


Jasstkn:
Configuration file for labeler can't be copied if you use shared workflow within the organization.
Is there any way to put configuration inside the workflow?

MaksimZhukov:
Hello @Jasstkn !
Here is statement from [Reusing workflows](https://docs.github.com/en/actions/using-workflows/reusing-workflows) GitHub documentation article: 

> If you reuse a workflow from a different repository, any actions in the called workflow run as if they were part of the caller workflow. For example, if the called workflow uses actions/checkout, the action checks out the contents of the repository that hosts the caller workflow, not the called workflow.

According to the documentation, your configuration file has to be in the same repository as caller workflow. Alternatively, you can clone the repository with the called workflow using the [checkout](https://github.com/actions/checkout) action:
```yml
- uses: actions/checkout@v3
  with:
    repository: 'Repository name with owner. For example, actions/checkout'
```

Please see the [GitHub documentation](https://docs.github.com/en/actions/using-workflows/reusing-workflows#using-inputs-and-secrets-in-a-reusable-workflow) for more examples.

MaksimZhukov:
I’m closing the issue. Please let us know if you have any questions.

jtmullen:
Hi @MaksimZhukov I have this same issue and I am confused by this solution you propose:

> According to the documentation, your configuration file has to be in the same repository as caller workflow. Alternatively, you can clone the repository with the called workflow using the [checkout](https://github.com/actions/checkout) action

It appears this action it hard-coded to make a request to the caller repo to get the labeler configuration. How does checking out the called repo (where my shared configuration lives) solve this?

lrstanley:
@jtmullen just FYI, https://github.com/actions/labeler/releases/tag/v4.3.0 / https://github.com/actions/labeler/pull/394 has added support for `configuration-path` to look at the local filesystem, before making the API call to the repo directly. This means you can clone the central repo down first, and directly reference it inside of the action.

Can see an example of how I'm accomplishing the same here (`.github` being the repo where I keep my config):

https://github.com/lrstanley/.github/blob/f163452238082e6c5af52edab262253cbaaaaf99/.github/workflows/triage.yml#L49-L66

Jaycon80:
Hello

Issue #16, opened 2019-08-30
[error]HttpError: Not Found


swuecho:
2019-08-30T04:47:25.3784277Z Current runner version: '2.157.2'
2019-08-30T04:47:25.3803572Z Prepare workflow directory
2019-08-30T04:47:25.4369293Z Prepare all required actions
2019-08-30T04:47:25.4675219Z Download action repository 'actions/labeler@v2'
2019-08-30T04:47:27.8427078Z ##[group]Run actions/labeler@v2
2019-08-30T04:47:27.8498581Z with:
2019-08-30T04:47:27.8498775Z   repo-token: ***
2019-08-30T04:47:27.8498811Z   configuration-path: .github/labeler.yml
2019-08-30T04:47:27.8498846Z ##[endgroup]
2019-08-30T04:47:28.4480335Z ##[error]HttpError: Not Found
2019-08-30T04:47:28.4807300Z ##[error]Not Found
2019-08-30T04:47:28.4832078Z ##[error]Node run failed with exit code 1
2019-08-30T04:47:28.5015714Z Cleaning up orphan processes

swuecho:
totally lost. Could anyone give me a hint? thanks

mansona:
I have hit this same issue when testing out the workflow 🤔 I don't know how public this is but here is the PR that is failing: https://github.com/mansona/actions-intro/pull/1

it is an empty repo that is just set up to test the action

nfadili:
Running into this as well. I've tried every combination of refs I can think of: 
- `uses: actions/labeler@v2` - 404
- `uses: actions/labeler@releases/v2` - 404
- `uses: actions/labeler@master` - Doesn't have a `lib` folder

gobadiah:
any news on this issue ?

rohan-buchner:
@gobadiah I just bumped my head on this... ensure you have both the following files
```
.github/workflows/labeler.yml   // with the action
.github/labeler.yml        //with your labels
```

gobadiah:
Thanks @rohan-buchner I already have the two files (I renamed the workflow from `label.yaml` to `labeler.yaml` just in case). The content of the workflow is copy pasted from the `README.md`, the content of the other file is just :

```
api: api/*
```

It was on a private repo, so I also tried doing this on a public repo, I get the same error.

frextrite:
I am facing the same issue
```
##[error]HttpError: Resource not accessible by integration
##[error]Resource not accessible by integration
[...]

PR #394, opened 2022-06-25
feat(config): support reading from local file if it exists

lrstanley:
Currently, **actions/labeler** can only read the configuration from the source repository, due to it using the context and API to fetch the configuration file. This PR adds support for first checking to see if a file matching the provided path exists locally, before it makes a call to the API to get the file.

The primary benefit for this, is it works well with reusable workflows, as all you have to do is grab the file from a central location, before running this on the target repository.

- closes https://github.com/actions/labeler/pull/141 (this PR is a simpler solution that allows for the same functionality)
- closes https://github.com/actions/labeler/issues/123


mac2000:
This change will also fix #16

[comment with back link](https://github.com/actions/labeler/issues/16#issuecomment-1465125506)

javiergonzalezGenially:
Any news about this? This is actually needed when labeler.yml is auto-generated in build time.

simbo:
@lrstanley I created a PR for your PR, that updates branch and uses async readFile.
https://github.com/lrstanley/labeler/pull/1

AndreiLobanovich (changes requested):
Hello, @lrstanley!
Thanks for your contribution!
I left a few comments, otherwise your solution looks great.

src/labeler.ts
```diff
@@ -96,10 +97,19 @@ async function getLabelGlobs(
...
-    configurationPath
-  );
+  let configurationContent: string;
+
+  try {
+    configurationContent = readFileSync(configurationPath, {
+      encoding: 'utf8'
+    });
+  } catch (error) {
+    core.debug(
```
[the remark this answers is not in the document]
lrstanley: Updated!

src/labeler.ts
```diff
@@ -96,10 +97,21 @@ async function getLabelGlobs(
...
+  let configurationContent: string;
+
+  try {
+    configurationContent = (
+      await readFile(configurationPath, {
+        encoding: 'utf8'
+      })
+    ).toString();
+  } catch (error) {
+    core.warning(
```
Youssef1313, line 109: I don't think this should be a warning.

src/labeler.ts
```diff
@@ -96,10 +97,21 @@ async function getLabelGlobs(
...
 ): Promise<Map<string, StringOrMatchConfig[]>> {
-  const configurationContent: string = await fetchContent(
-    client,
-    configurationPath
-  );
+  let configurationContent: string;
+
+  try {
+    configurationContent = (
+      await readFile(configurationPath, {
```
Youssef1313, line 104: Documentation on proper usage should be added.

src/labeler.ts
AndreiLobanovich: Thank you for your comment on this, @Youssef1313. We discussed this in our team and came up to the conclusion, that making this message a warning makes output too noisy. Instead the following could de a more suitable solution
`const warningPrefix = '[warning]';
core.info(`${warningPrefix} the rest of the message`);`
@lrstanley, what do you think?

src/labeler.ts
AndreiLobanovich: Sure. @lrstanley, could you also update README accordingly?

src/labeler.ts
lrstanley: I do agree that showing warnings for this can be a bit noisy. Decided on the approach you've mentioned. Not sure if `warningPrefix` should be defined outside of `labeler.ts`, but that's where I've stuck it for now.

README.md
```diff
@@ -112,7 +112,7 @@ Various inputs are defined in [`action.yml`](action.yml) to let you configure th
 | Name | Description | Default |
 | - | - | - |
 | `repo-token` | Token to use to authorize label changes. Typically the GITHUB_TOKEN secret, with `contents:read` and `pull-requests:write` access | `github.token` |
-| `configuration-path` | The path to the label configuration file | `.github/labeler.yml` |
+| `configuration-path` | The path to the label configuration file. If not found in the current working directory, will read from the source repository via the Github API. | `.github/labeler.yml` |
```
AndreiLobanovich, line 115: By default action searches for configuration in .github folder. We should better mention this instead of `working directory`.

AndreiLobanovich (changes requested):
Hey, @lrstanley! Thanks for your changes. Could you also provide us with the test case for this functionality?

README.md
Youssef1313, line 115: Can there be a "full" usage example for the new feature? How the labeler.yml will be in the current directory and/or `.github`?

AndreiLobanovich:
Hey, @lrstanley! 
Do you have some spare time to finish this PR? Is there anything I can help you with?

README.md
lrstanley: `working directory` would be correct, though? If `configuration-path` is not specified, the chart will read from `.github/labeler.yml`, though, and that's already documented in the README.

lrstanley: Honestly don't see how that's necessary. It's a path to a file, why would that need an example?

Youssef1313: To give an idea to the reader how they would use this feature. As a reader I'll be wondering: How would I get the configuration file in the machine running the workflow? Is there a best practice to that matter?

lrstanley:
> Hey, @lrstanley! Thanks for your changes. Could you also provide us with the test case for this functionality?

I don't see any existing mock/testing functionality for the actual reads from the API, and if I write a test, it should probably cover the case if a file doesn't exist (or similar), which means it'd try and read from the remote. Not sure best practice on how to implement that. Do the octokit libraries provide ways of mocking those calls?

src/labeler.ts
```diff
@@ -97,10 +100,21 @@ async function getLabelGlobs(
...
+        encoding: 'utf8'
+      })
+    ).toString();
+  } catch (error) {
+    core.info(
+      `${warningPrefix} configuration file (path: ${configurationPath}) not found locally (${error}), fetching via the api`
```
Youssef1313, line 113: I wouldn't consider this as a warning.

Youssef1313: @AndreiLobanovich The warning prefix makes this look stronger than it should IMO, even if logged with info severity. Is it necessary to have the warning prefix?

AndreiLobanovich: @Youssef1313, the thing is - users were having a hard time understanding what's wrong when config file was misconfigured. info-warning should address that.

README.md
AndreiLobanovich: @lrstanley `working directory` might be misleading as it stands for the root directory of the project. 
`If file is not found at the provided path to config file, will read from the source repository via the GitHub API` - is what we came up with.

AndreiLobanovich:
@lrstanley as for tests - you may utilize the following [approach](https://github.com/actions/setup-node/blob/main/__tests__/main.test.ts#L239).

src/labeler.ts
MaksimZhukov: @lrstanley could you please pull the latest changes from the `main` branch? We merged the [pull request](https://github.com/actions/labeler/pull/475) that improves error message handling. With these changes, the warning prefix is no longer needed so please remove it.

MaksimZhukov, line 113: After you pull the latest changes from the `main` branch, please move `core.info` to the `try` block and add a check to make sure if the configuration file exists. If it exists, let's read its content, if not, then print a message and try to fetch it via API.

MaksimZhukov:
> I don't see any existing mock/testing functionality for the actual reads from the API, and if I write a test, it should probably cover the case if a file doesn't exist (or similar), which means it'd try and read from the remote. Not sure best practice on how to implement that. Do the octokit libraries provide ways of mocking those calls?

I think we can add two tests here:
1) `uses local configuration file if it exists`
2) `fetches the content of configuration file via API if not found locally`

Here is an example of the first test case: 
```ts
  it('uses local configuration file if it exists', async () => {
    const configFile = 'only_pdfs.yml';
    const configFilePath = path.join(__dirname, 'fixtures', configFile);
    
    configureInput({
      'configuration-path': configFilePath
    });

    await run();

    expect(reposMock).toHaveBeenCalledTimes(0);
  });
```

AndreiLobanovich (approved):

MaksimZhukov (approved):

dmitry-shibanov (approved):

MaksimZhukov:
@lrstanley thank you for the contribution!

PR #349, opened 2022-04-19
Added ability to pass in an optional PR number as a parameter

credfeto:
So so it can be used in a workflow that creates a PR automatically like `repo-sync/pull-request` then add in labels at creation rather than having to run an additional workflow

```yaml
name: Auto PR Creation
on:
  push:
    branches-ignore:
    - master
    - main

jobs:
  create-pull-request:
    runs-on: ubuntu-latest

    steps:
    - uses: actions/checkout@v3
      with:
        fetch-depth: 1
        token: ${{ secrets.GITHUB_TOKEN }}
    - run: |
        echo COMMIT_MSG=`git log -1 --pretty=%B` >> $GITHUB_ENV

    - id: open-pr
      uses: repo-sync/pull-request@v2.6
      with:
        source_branch: ""                     # If blank, default: triggered branch
        destination_branch: "main"            # If blank, default: master
        pr_assignee: "${{ github.actor }}"    # Comma-separated list (no spaces)
        pr_label: "auto-pr"                   # Comma-separated list (no spaces)
        pr_draft: true                        # Creates pull request as draft
        pr_title: "${{ env.COMMIT_MSG }}"
        pr_body: "${{ steps.pr-template.outputs.content }}"
        github_token: ${{ secrets.GITHUB_TOKEN }}

    - name: Details
      run: |
          echo "URL: ${{steps.open-pr.outputs.pr_url}}"
          echo "PR: ${{steps.open-pr.outputs.pr_number}}"
          echo "CF: ${{steps.open-pr.outputs.has_changed_files}}"

    - uses: credfeto/labeler@feature/pass-in-pull-request-or-issue
      with:
        repo-token: "${{ secrets.GITHUB_TOKEN }}"
        configuration-path: .github/labeler.yml
        sync-labels: true
        pr-number: ${{steps.open-pr.outputs.pr_number}}
```

Also would fix https://github.com/actions/labeler/issues/304

PR #411, opened 2022-08-02
Add a new `pr-number` input

Andarist:
**Motivation**

It's useful when a previous step/job in a workflow creates a PR, or in general when the workflow is executed in the context of a different event (could be a repository dispatch or just anything) and when the PR number might come from an arbitrary source. 

I had a feature request to add `labels` input to the https://github.com/changesets/action (this action creates/updates a PR) but I think that it would be much better to achieve this through composition. It seems to me that people who needs it could use the output of my action as input to this one.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,6 +116,7 @@ Various inputs are defined in [`action.yml`](action.yml) to let you configure th
 | `repo-token` | Token to use to authorize label changes. Typically the GITHUB_TOKEN secret, with `contents:read` and `pull-requests:write` access | N/A |
 | `configuration-path` | The path to the label configuration file | `.github/labeler.yml` |
 | `sync-labels` | Whether or not to remove labels when matching files are reverted or no longer changed by the PR | `false`
+| `pr-number` | Pull request number to label instead of the inferred number from the context |
 
 # Contributions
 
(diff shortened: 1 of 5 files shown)


pje:
Thanks for your contribution, @Andarist! 

This is exactly the kind of PR I love to see:

- flexible, clear use-case
- limited in scope
- has tests

Please fix the conflicts in `dist/index.js` (by regenerating it) and I'm happy to merge.

Andarist:
@pje done

Andarist:
@pje I've rebased this again - would you find some time to land this?

shaunofneuron:
I'm pretty sure others will figure out how to run this manually using the `gh cli` but it would be helpful once merged if there were some future guide/doc on how to trigger with this new parameter? Perhaps something like?

```bash
echo '{"pull_request":1234}' | gh workflow run labeler.yaml --json
```
I came here because I was looking to trigger against PR's that need labels, even adding in the `workflow_dispatch` trigger so do so.  Then I realized this wasn't possible since "PR" is only derived from the context.

MaksimZhukov:
Hello @Andarist! Thank you for your willingness to contribute and improve the action!
There is a [pull request](https://github.com/actions/labeler/pull/349) that was opened before yours, and we are going to accept it.
I'm closing it as a duplicate.

PR #426, opened 2022-08-17
 Add optional input for selectively opting-out of sync

amccall-eigt:
New Input: `sync-ignore-labels`
Required: No

Use Case: We would like to be able to add specific labels at PR creation that do not interact with the labeler labels. This is particularly useful for opting out of build automation that runs based on assigned labels.

Test case has been added to validate the behavior and all test cases are passing.

README.md has been updated to reflect new input.


MaksimZhukov:
Hello @amccall-eigt! 
Could you please describe your use case in a little more detail? 
My understanding is that you would like to add some labels without using the `labeler` action and don't want the action to remove them if they don't match the globs. The action only manages the labels that are specified in the configuration file. If you want to manage these labels yourself, why do you need to add them to the labeler config file? If you want to add labels in two ways and the action removes them, maybe it is better to change the glob patterns for those labels or create separate labels that will suit your needs? It seems to me that you have labels with an ambiguous non-obvious meaning.

MaksimZhukov:
I am closing this PR. Please let us know if you have any questions or concerns

PR #456, opened 2022-11-15
Early exit when no files are changed.

nathanhammond:
As a consequence of the `checkAll` function call returning `true` if the length of `changedFiles` is 0, this must early-exit in order to avoid labeling empty PRs.

Example labeling failure here: https://github.com/vercel/turbo/pull/2706


nathanhammond:
The source of the error is here:
https://github.com/actions/labeler/blob/5c7539237e04b714afd8ad9b4aed733815b9fab4/src/labeler.ts#L202-L214

The loop on lines 205-210 does not execute and therefore the `checkAll` check returns `true`.

nathanhammond:
Rebased and updated to match new code style. As all changes were formatting, the `dist` asset didn't change.

src/labeler.ts
```diff
@@ -40,6 +40,23 @@ export async function run() {
 
     const labels: string[] = [];
     const labelsToRemove: string[] = [];
+
+    if (changedFiles.length === 0) {
```
dmitry-shibanov, line 44: could you please change it to if (!changedFiles.length) {

src/labeler.ts
```diff
@@ -40,6 +40,23 @@ export async function run() {
...
+      core.debug(`exiting early because pr #${prNumber} has no changed files.`);
+
+      if (syncLabels) {
+        for (let label of pullRequest.labels) {
+          if (label.name !== undefined) {
```
dmitry-shibanov, line 49: I think we can change it to the same check if (label.name) {

nathanhammond: Confirmed that name can't be blank at creation time. (returns 422)

`{"name":["Name can't be blank"],"message":"Name can't be blank"}`

nathanhammond:
Updated and rebuilt.

AndreiLobanovich:
Hello @nathanhammond! Could you please tell me, how do you create an empty PR? The link you provided as a labeler failure leads to a pull request. If you could provide the link with the run in which labeler failed - that would help me investigate the issue.

src/labeler.ts
```diff
@@ -40,6 +40,23 @@ export async function run() {
 
     const labels: string[] = [];
     const labelsToRemove: string[] = [];
+
```
AndreiLobanovich, line 43: Could you please move this right after the following line?
`const changedFiles: string[] = await getChangedFiles(client, prNumber);`

src/labeler.ts
```diff
@@ -40,6 +40,23 @@ export async function run() {
...
+    if (!changedFiles.length) {
+      core.debug(`exiting early because pr #${prNumber} has no changed files.`);
```
AndreiLobanovich, line 45: This should be a warning. Creating an empty pr should definitely be noticed by the user.

src/labeler.ts
```diff
@@ -40,6 +40,23 @@ export async function run() {
...
+
+      if (syncLabels) {
```
AndreiLobanovich, line 47: There is no need to remove labels, because early exit doesn't allow them to appear.

AndreiLobanovich (changes requested):
Hello, @nathanhammond! 
Thanks for your contribution and attention to details! I managed to reproduce the incorrect behavior. Could you please make a few minor changes?

AndreiLobanovich:
Hello, @nathanhammond!

Didn't hear from you for a while. Do you have some spare time to finish work on this PR? Is there anything I can help you with?

AndreiLobanovich:
Hey, @nathanhammond, we miss you a lot. If you could find some spare time to finish work on this PR that would be great. 
Hope you are doing well.

AndreiLobanovich (approved):

IvanZosimov (approved):

MaksimZhukov (approved):

dmitry-shibanov (approved):

MaksimZhukov:
@nathanhammond thank you for the contribution!

PR #468, opened 2022-12-13
chore: Create check-dist.yml

jongwooo:
Signed-off-by: jongwooo <jongwooo.han@gmail.com>

## Description

This adds a workflow to check the contents of the checked-in `dist/index.js` against the expected version.

See https://github.com/actions/upload-artifact/pull/227 for background and rationale.


jongwooo:
Duplicated #463

PR #487, opened 2023-01-20
Add the allofany matching type (#423)

dfandrich:
This is true when all files match any of the patterns.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,13 +26,15 @@ For more control over matching, you can provide a match object instead of a simp
 ```yml
 - any: ['list', 'of', 'globs']
   all: ['list', 'of', 'globs']
+  allofany: ['list', 'of', 'globs']
 ```
 
-One or both fields can be provided for fine-grained matching. Unlike the top-level list, the list of path globs provided to `any` and `all` must ALL match against a path for the label to be applied.
+One two or three fields can be provided for fine-grained matching. Unlike the top-level list, the list of path globs provided to `any` and `all` must ALL match against a path for the label to be applied.
 
 The fields are defined as follows:
 * `any`: match ALL globs against ANY changed path
 * `all`: match ALL globs against ALL changed paths
+* `allofany`: match ANY globs against ALL changed paths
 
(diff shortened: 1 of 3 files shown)


hstoresin:
npm run verify -- --tree=merkle-tree.json --hash=08a72001307d26afe079f6582ea5aacbeb7e1a01

dfandrich:
It would be nice to get this into v5. I count 3 open issues that this feature would address, plus my own use case.

silverwind:
I would call this matcher `only`, e.g. it only matches when all specified globs changed exactly with no unmatched files present, like in this table:

| change files | a | a,b | a,b,c |
| ------------ | - | --- | ----- |
| all: [a, b]  | ❌ | ✅   | ✅     |
| only: [a, b] | ❌ | ✅   | ❌     |
| any: [a, b]  | ✅ | ✅   | ✅     |

Is it how this works?

dfandrich:
I don't think that table is accurate, if I'm reading it correctly (where each glob [a, b] matches a single, different file). If I make the globs look like globs for clarity of this interpretation, the table would be:

| changed files |	a |	a,b |	a,b,c |
|--|--|--|--|
| all: [a*, b*] | ❌ |❌ |❌ |
| allofany: [a*, b*] |	✅ |	✅ 	|❌ |
| any: [a*, b*] |	❌ |	✅ |	✅ |

The `all` case can never match anything because a* and b* are disjunct globs, so maybe I misinterpreted your version of the table.

While I'm not married to the name, it should be consistent with the existing operators. `any` and `all` refer to how many paths must match the globs, whereas `only` doesn't fit that pattern since it doesn't say anything about the paths.

dfandrich:
This PR has been superseded by #680.

brianluckham:
dfandrich:allofany

dfandrich:
Note that the semantics of PR #680 turn out not to match this PR's. `AnyGlobMatchesAllFiles` is not the same as `allofany`. It sounds like it should be, but `AnyGlobMatchesAllFiles` has the semantics of `OneGlobMatchesAllFiles` instead.

Issue #399, opened 2022-07-12
Labels not applied on PRs FROM forks


gregsdennis:
This is a bit related to #12 and #36, except that it's the other way around.

I have a repo (https://github.com/gregsdennis/json-everything), and the labeler always fails on PRs submitted by others ([example](https://github.com/gregsdennis/json-everything/pull/296)).

I expect that since the PR is in the primary repo, not the fork, it's should be getting my token (which works on all of _my_ PRs) and running the action with that.

BitsOfAByte:
Can replicate this issue, all PRs coming into my repository from forks fail every time with `
HttpError: Resource not accessible by integration`

Hans5958:
For these cases, `pull_request_target` should be used compared to `pull_request`.

gregsdennis:
Sorry.  I'm finally getting around to updating this as I finally have an [external PR](https://github.com/gregsdennis/json-everything/pull/369).

The change to `pull_request_target` isn't working for me.  I even added the explicit permissions as you have in the README.

gregsdennis:
Nevermind.  That example was a rebasing issue.  All is well.  Thanks for the pointers.

remyleone:
I have an issue in this run: https://github.com/scaleway/terraform-provider-scaleway/actions/runs/17070268970/job/48397094646?pr=3269#step:3:9 
While having:

```
name: "Pull Request Labeler"
on:
  - pull_request_target

jobs:
  labeler:
    permissions:
      contents: read
      pull-requests: write
    runs-on: ubuntu-latest
    steps:
      - name: Checkout
        uses: actions/checkout@v4
      - name: Labeler
        uses: actions/labeler@v5
```

PR #491, opened 2023-01-25
Warn about the limitations of GitHub tokens

dfandrich:
The action will return

  Error: HttpError: Resource not accessible by integration

when triggered on a PR on a forked repository.

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,6 +131,24 @@ label1:
 - path/to/folder/**
 ```
 
+## Permissions
+
+In order to add labels to pull requests, the GitHub labeler action requires
+write permissions on the pull-request. However, when the action runs on a pull
+request from a forked repository, GitHub only grants read access tokens for
+pull_request events, at most. If you encounter an Error: HttpError: Resource
+not accessible by integration, it's likely due to these permission constraints.
+To resolve this issue, you can modify the `on:` section of your workflow to use
+[pull_request_target](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target)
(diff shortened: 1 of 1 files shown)


dfandrich:
I just discovered issue #399, so this warning should be updated.

AndreiLobanovich:
Hey, @dfandrich! Thanks for your contribution! We decided to go with your approach - adding permissions section in README. However we have updated the text. 

In order to add labels to pull requests, the GitHub labeler action requires write permissions on the pull-request. However, when the action runs on a pull request from a forked repository, GitHub only grants read access tokens for pull_request events, at most. If you encounter an Error: HttpError: Resource not accessible by integration, it's likely due to these permission constraints.
To resolve this issue, you can modify the `on:` section of your workflow to use [pull_request_target](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target) instead of `pull_request`  (see example [above](#create-workflow)). This change allows the action to have write access, because pull_request_target alters the [context of the action](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target) and safely grants additional permissions.
Refer to the [GitHub token permissions documentation](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token) for more details about access levels and event contexts.

Could you please change the text like that?

dfandrich:
Done.

MaksimZhukov (approved):

AndreiLobanovich (approved):

IvanZosimov (approved):

Issue #115, opened 2020-11-22
Issues cannot have more than 100 labels


Nuru:
I have auto-labeler set up on a monorepo and sometimes it generates (per configuration) more than 100 labels. However, it then fails with the error:
```
HttpError: Validation Failed. Issues cannot have more than 100 labels: 
```
There needs to be some mechanism to handle the case when it would generate more than 100 labels. Ideas:
- Some way to truncate the list (and perhaps add a "truncated" label)
- A way to prioritize labels, so that the most important labels are set even if some others are truncated
- Some way to merge labels if there are too many
- An alternative label configuration to use if the first one generates too many labels.

radulle:
As there are more and more mono-repos this issue will come up more and more so it needs to be resolved.

PR #497, opened 2023-02-02
fix: Limit number of labels added to 100

markmssd:
**Description:**
Github Issues can have at most 100 labels assigned to them. This PR limits the labels being added to 100, and logs the rest as a warning.

Closes https://github.com/actions/labeler/issues/561
Closes https://github.com/actions/labeler/pull/562

~~**TODO: Unit tests, will add them if the PR makes sense!**~~

**Related issue:**
[Add link to the related issue.](https://github.com/actions/labeler/issues/115)

**Check list:**


dist/index.js
```diff
@@ -66,20 +68,29 @@ function run() {
             const labelGlobs = yield getLabelGlobs(client, configPath);
             const labels = [];
             const labelsToRemove = [];
+            const excessLabels = [];
             for (const [label, globs] of labelGlobs.entries()) {
                 core.debug(`processing ${label}`);
                 if (checkGlobs(changedFiles, globs)) {
-                    labels.push(label);
+                    if (labels.length >= GITHUB_MAX_LABELS) {
```
macksal, line 75: I think this neglects the number of labels that are already present. Say the PR has 50 labels, and we come up with 100 that we want to apply, but only 25 of those are already present. The correct result would be to add 50 new ones, 25 remain the same, and 25 would be excess. But the current code reports 100 labels with no excess.

markmssd: Good point. I've moved the `addLabels` to happen AFTER the `removeLabels` (which happens with `syncLabels` enabled only) to mitigate this issue. But maybe there's a better way still, I'll check again. Thanks!

macksal: To me, an easy solution seems to be first getting the labels already on the issue with [`listLabelsOnIssue`](https://octokit.github.io/rest.js/v19#issues-list-labels-on-issue). Then, you have all the information needed to ensure the _total_ count remains less than 100 while tracking accurately all additions/removals that you want to do.

markmssd:
@macksal Alright so I ended up simply using `pullRequest.labels`, which already holds all the PR's labels. Then:

- I hydrate the `labels` array with those current labels only if `syncLabels=false`
- Push all matching labels into the `labels` array (making sure there is no duplicates)
- Splice the array at 100 (the maximum number of labels allowed on a PR). This splits the original `labels` array to a length of 100, and all the rest is returned into `excessLabels` array.
- If `syncLabels=true`, I use `setLabels`, otherwise I `addLabels`. Either way, it's guaranteed to never set/add more than 100 labels
- If there was an excess in `excessLabels` array, simply log them

If that makes sense I'll add unit tests!

markmssd:
@macksal I've rebased my branch to resolve conflicts. Kindly let me know if it needs any more changes (other than unit tests)!

dusan-trickovic:
Hello, @markmssd ! Thanks for creating the PR - we love and endorse customer contributions! Overall, your PR looks good, it definitely deals well with the scenario of trying to add more than 100 labels and with what to do with the excess.

However, we would also like you to take a look at [this PR](https://github.com/actions/labeler/pull/562) and its changes. Maybe it can help you see if anything could be added and/or refactored. Furthermore, those two together could solve [this issue](https://github.com/actions/labeler/issues/561), as the latter deals with reducing the array by filtering duplicates, but not limiting the number of labels to 100, which former takes into account.

In addition to that, there's another finding that could prove to be beneficial to the solution as well and which we strongly recommend incorporating into the final application. Namely, upon some inspection on our end, we have discovered that the Action will likely keep throwing `HttpError: Server Error` unless we split the total number of labels in the array past the certain index and send them in parts instead. That 'golden number', as far as our tests go, turns out to be 48.

What that means is that if there are less than 48 labels present in the array, those can be sent all at once without any problems. As soon as that number is exceeded, the array should be split. We suggest splitting it in half, as this seems to be the most efficient way. 

However, keep in mind that, as the number of labels keep increasing (e.g. 100 labels), splitting it in half could also give us more than 48 labels to work with. As a solution to this, you could put some kind of a loop in place that could perform the action above for as long as needed (e.g. 100 gets split into 2x50, which then split to 2x25).

Example of this part of the logic:

```
const LABELS_LIMIT_TO_ADD_AT_ONCE = 48;

while (labelsToBeAdded.length > 0) {
    let splicedLabelsList: string[] = [];
    splicedLabelsList = labelsToBeAdded.splice(0, LABELS_LIMIT_TO_ADD_AT_ONCE / 2);
    await sendLabels(client, prNumber, splicedLabelsList);
}
```

I hope I explained this clearly :) For any further clarifications, questions and suggestions, please feel free to reach out to us again and we will try to answer them as best as we can. And, of course, feel free to also add some unit tests :)

Thank you for your time and contribution to our software :)

dusan-trickovic:
Hello, @markmssd ! I just wanted to give you a little ping to see if there are any new updates on your end? Maybe some thoughts about my previous suggestion? Thank you in advance :)

markmssd:
Hey @dusan-trickovic , thanks for the thorough explanation!

I just got back from vacation and I'll need a few days before getting back to this PR.

So if I understand correctly, this HTTP Server Error happens only when using the "addLabels" method, or also when using "setLabels"? If that's the case, I could also change the PR to ONLY use setLabels, and it would fix all issues.

The logic would be:
- if syncLabels=true, use setLabels with new labels only.
- if syncLabels=false, use setLabels with current PR's labels + new ones

However if the HTTP Server Error happens with setLabels too, then I'll go with your suggestion with the golden number 48, what do you think?

src/labeler.ts
```diff
@@ -24,7 +28,7 @@ export async function run() {
       return;
     }
 
-    const client: ClientType = github.getOctokit(token);
+    const client: ClientType = github.getOctokit(token, {}, pluginRetry.retry);
```
markmssd, line 31: So I've noticed that when using the [setLabels API](https://docs.github.com/en/rest/issues/labels?apiVersion=2022-11-28#set-labels-for-an-issue), it would still fail with 5xx error if adding >50 labels (or if there's currently >50 labels and we're trying to remove them all), but retrying the query then passes.

Thus I'm simply adding the `retry` plugin instead. I think it's a good thing to add even if that issue gets officially fixed by Github on the backend, for resilience.

dusan-trickovic (changes requested):
Hello @markmssd ! Thank you again for creating this PR! I have written a few initial review points that should be addressed before continuing on. Please let us know if you have any additional questions about them, any additional suggestions or concerns :)

src/labeler.ts
```diff
@@ -11,11 +12,14 @@ interface MatchConfig {
...
 
+// Github Issues cannot have more than 100 labels
+const GITHUB_MAX_LABELS = 100;
+
 export async function run() {
   try {
     const token = core.getInput('repo-token');
     const configPath = core.getInput('configuration-path', {required: true});
-    const syncLabels = !!core.getInput('sync-labels');
+    const syncLabels = core.getBooleanInput('sync-labels');
```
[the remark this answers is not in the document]
markmssd: Ah good to know, I'll revert the change.

src/labeler.ts
```diff
@@ -243,32 +249,15 @@ function checkMatch(
   return true;
 }
 
-async function addLabels(
```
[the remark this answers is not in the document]
markmssd: You're absolutely right, good catch! I've modified the code slightly so it preserves the manually added labels regardless of `sync-labels` (and adjusted the tests to properly test this use case).

src/labeler.ts
```diff
@@ -243,32 +257,22 @@ function checkMatch(
...
+function removeLabel(labels: string[], label: string): void {
```
MaksimZhukov, line 260: The current name confuses me a little. Based on its name, I think it still makes an API call. Can we make this more specific?
```suggestion
function removeLabelFromList(labels: string[], label: string): void {
```

MaksimZhukov (commented):
Hello @markmssd! Thank you for the contribution!
Looks good to me! I left a minor comment

src/labeler.ts
markmssd: Makes sense, done!

MaksimZhukov (approved):

dmitry-shibanov (approved):

dusan-trickovic (approved):

marko-zivic-93 (approved):

MaksimZhukov:
@markmssd could you please resolve conflicts?

src/labeler.ts
```diff
@@ -39,23 +43,33 @@ export async function run() {
       configPath
     );
 
-    const labels: string[] = [];
-    const labelsToRemove: string[] = [];
+    const labels: string[] = pullRequest.labels.map(label => label.name);
```
dsame, line 46: May be it worth to create Set object at this point and avoid `!labels.includes(label)` check further

dsame (approved):

.gitignore
```diff
@@ -1,3 +1,5 @@
 .DS_Store
 node_modules/
-lib/
\ No newline at end of file
+lib/
+.vscode/
+.idea/
```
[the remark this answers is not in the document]
markmssd: Oh I can remove them haha, it was just getting annoying locally

src/labeler.ts
markmssd: I initially tried with a Set but the code was getting really ugly, I'll try that again 🤔

markmssd: Alright I pushed changes that use Sets instead, let me know what you think

markmssd:
Conflicts have been served ✅ I also added an additional check to only call `setLabels` if there was actually a change, to avoid unnecessary API calls, and fixed the tests accordingly.

MaksimZhukov (approved):

marko-zivic-93 (approved):

IvanZosimov (approved):

dusan-trickovic:
Hello @markmssd ! The PR looks good, we will be merging it now. Thank you very much for your contribution! :)

asdaiipeyman-svg:
[magisk_log_2025-10-17T03.44.05.log](https://github.com/user-attachments/files/22961819/magisk_log_2025-10-17T03.44.05.log)

asdaiipeyman-svg (commented):
2025-10-06T12:58:38Z
2025-10-06T12:58:39Z @markmssd #497 #908 #905 #713

PR #408, opened 2022-07-27
Adds `outputs.labels`

daczarne:
Exposes the added labels as an output (string). If no labels where added, it returns an empty string.

Merging this pull request should close #407.

PR #60, opened 2020-03-24
Added output

danielsht86:
In order to use this action as one step in a series of steps in a workflow (and use the changed labels as input to downstream actions), this PR outputs the added labels and all labels as output to the action.

```
outputs:
  new-labels:
    description: 'List of all new labels'
  all-labels:
    description: 'List of all labels'
```

The only changes here are the new definition of the output and piping the new labels and all labels (taken from the result of the API call) and sending them as output via `core.setOutput`

PR #586, opened 2023-06-13
add action outputs

caioaao:
**Description:**
Add two outputs to the action: `labels` and `removed-labels`. 

**Related issue:**
Closes #408 (previous PR that didn't include `labels-to-remove`)
Closes  #60 

**Check list:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,6 +131,13 @@ label1:
 - path/to/folder/**
 ```
 
+#### Outputs
+
+| Name     | Description | Default |
+| -        | -           | - |
+| `labels` | A string containing the labels added, separated by white spaces. If no labels where added, it returns an empty string (`""`) | N/A |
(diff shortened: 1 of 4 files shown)


MaksimZhukov:
Hello @caioaao ! Thank you for the PR!
We are going to accept #60. It adds the `all-labels` output which is very useful because the `github` context contains a set of labels that are only available at the time the event is triggered, changes made during the workflow run won't show up there. 
Could you please describe a use case for the `removed-labels` output?

caioaao:
Hi @MaksimZhukov ! Sounds good. I didn't have any specific use-case in mind, just added it for completeness sake and I think having the `all-labels` output already fulfills my needs too :) . Since the other PR seems to be moving forward, I'll close this one. Thanks! :wave:

PR #611, opened 2023-07-11
Make `jest-each` a devDependency

ericcornelissen:
**Description:**
Change the dependency on `jest-each` from a regular dependency to a development dependency (devDependency) as it is only used in tests.

**Related issue:**
This relates to #394 which introduced `jest-each` as a dependency.

**Check list:**


ericcornelissen:
As a second suggestion, the functionality of `jest-each` has been part of the Jest core since v23 ([ref](https://jestjs.io/blog/2018/05/29/jest-23-blazing-fast-delightful-testing#jest-each)). For the [version of Jest used in this project](https://github.com/actions/labeler/blob/ac9175f8a1f3625fd0d4fb234536d26811351594/package.json#L46) (as well as later versions) it can be used as [`test.each`](https://jestjs.io/docs/27.x/api#testeachtablename-fn-timeout).

I did not see the decision to use the package `jest-each` over the built-in `test.each` function being discussed in #394. I'd propose to make that switch, if you agree I can update this Pull Request accordingly.

As a reference, the diff would look something like (in addition to removing the dependency on `jest-each` altogether):

```diff
  import path from 'path';
  import fs from 'fs';
- import each from 'jest-each';

  jest.mock('@actions/core');

...

    });

-   each([
+   test.each([
      [new HttpError('Error message')],
      [new NotFound('Error message')]
-   ]).test(
+   ])(
      'should warn if configuration file could not be fetched through the API, log error and fail the action',
      async error => {
```

IvanZosimov:
Hi, @ericcornelissen 👋 Thank you for noticing this! For sure, please, switch the code to use the built-in `test.each`. Moreover, in that case, `jest-each` can be removed from the list of dependencies completely.

ericcornelissen:
> For sure, please, switch the code to use the built-in test.each. Moreover, in that case, jest-each can be removed from the list of dependencies completely.

Done and done 🙂

MaksimZhukov (approved):

IvanZosimov (approved):

dmitry-shibanov (approved):

PR #688, opened 2023-09-27
feat(match): matches by PR title and body

kalvarez2:
**Description:**
Enhances the glob matchers to support matching a text on PR title or body. 
The change is backwards compatible, no need to change the existing configurations. Any and All continue to work.
The matchers now support prefix "title:" and "body:"
Simple example:

label1:
- "body:#potato"

=>Would add the label "label1" if the body contains the string #potato

The existing all and any continue to work:
isText:
- all: ['*.txt', 'body:#isOnlyText']

=> Would set the label isText if both all the files have extension txt, and the body contains "#isOnlyText"

Documentation changes are encouraged, but because this is backwards compatible, are not "absolutely required".

**Related issue:**
https://github.com/actions/labeler/issues/55

**Check list:**


jmeridth:
@kalvarez2 any chance we can pick this back up?  Would love this feature to be available in this action. @MaksimZhukov is this something this action is interested in?

mofosyne:
Would like to follow up on this as an alternative to using https://github.com/Bhacaz/label-regex

mwarkentin:
This would be useful for us as well, using branch names can catch **some** of the PRs I'd like to label, but I suspect we'll miss a bunch which might have a keyword we could use in the PR description or title.

My use case is to label incident-related PRs.

PR #714, opened 2023-12-08
any-glob-all-files: combine glob outputs to match all files on all globs

jnewb1:
checkout the discussion at https://github.com/actions/labeler/issues/423

there is currently no way to apply the "CI" label with the following rule

```any-glob-all-files: [Jenkinsfile, .github/**]```

when a PR which changes *both* Jenkinsfile and github actions files

essentially, this test fails, which to me makes "any glob matches all files" extremely confusing.
```
describe('check issue #423', () => {
  const changedFiles = ['test.md', 'test.puml'];

  describe('when at least one pattern matches all files', () => {
    const globPatterns = ['*.md', '*.puml'];

    it('returns true', () => {
      const result = checkIfAnyGlobMatchesAllFiles(
        changedFiles,
        globPatterns,
        false
      );
      expect(result).toBe(true);
    });
  });
});
```

perhaps this behavior should be moved to another rule, since it also breaks this potential use case: https://github.com/actions/labeler/blob/8558fd74291d67161a8a78ce36a881fa63b766a9/__tests__/changedFiles.test.ts#L228C3-L239C6

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,4 @@
+{
+	"name": "Node.js",
+	"image": "mcr.microsoft.com/devcontainers/javascript-node:1-20-bullseye"
+}
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,4 +1,0 @@
-{
-	"name": "Node.js",
-	"image": "mcr.microsoft.com/devcontainers/javascript-node:1-20-bullseye"
-}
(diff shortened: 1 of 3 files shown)


dfandrich:
As mentioned in #423, I'd rather fix `any-glob-to-all-files` with these new semantics, since I'm pretty sure its current implementation is inadvertently implemented incorrectly.

jnewb1:
this would be extremely helpful in our repo to get this merged!

jnewb1:
@MaksimZhukov

jnewb1:
who do I ask for a review to get this merged?

Issue #731, opened 2023-12-20
any-glob-all-files does not working intuitevely


jnewb1:
there is currently no way to apply the "CI" label with the following rule

any-glob-all-files: [Jenkinsfile, .github/**]

when a PR which changes both Jenkinsfile and github actions files

open PR to fix: https://github.com/actions/labeler/pull/714

IvanZosimov:
Hi, @jnewb1 👋 Thanks for the issue and PR, we will take a look.

IvanZosimov:
Hi, @jnewb1, I believe that for your particular use case the glob's syntax of grouping conditions fits well. What I mean by that is:
```
any-glob-all-files: "{Jenkinsfile,.github/**}"
```
Could you, please, check that logic of this grouping is what you want to achieve and ping me back?

jnewb1:
> Hi, @jnewb1, I believe that for your particular use case the glob's syntax of grouping conditions: `{a,b,c}` fits well. What I mean by that is:
> 
> ```
> 
> any-glob-all-files: "{Jenkinsfile,.github/**}"
> 
> ```
> 
> Could you, please, check that logic of this grouping is what you want to achieve and ping me back?

Thank you, this syntax works for my use case, though I think it is less readable

dfandrich:
That that {} syntax documented anywhere? I couldn't find it. It's certainly not intuitive, and `any-glob-all-files` is still misnamed; it should be `one-glob-all-files`.

PR #733, opened 2023-12-29
Add test and documentation for any-glob-to-all-files

dfandrich:
This test verifies that the glob brace syntax will match any glob
against all files. Document that any-glob-to-all-files means any single
glob. This reduces the ambiguity that caused #731 and other issues to be
opened. Also, fix an any-glob-to-any-file example so it actually matches
within the entire repository.


IvanZosimov:
Hi, @dfandrich 👋 Thanks for the PR, we will take a look!

IvanZosimov (approved):

dusan-trickovic (approved):

PR #109, opened 2020-10-22
enable to add label by title

nakamasato:

PR #866, opened 2025-04-29
feat: add pr title matching

### LINKED ISSUE #809: Support for labeling based on PR title
**Description:**
It would be nice if we could use this action to set up a step that would label PRs based on their title.
Take `feat(xxx): something` and `fix(yyy): something else` as examples.
We can write a configuration like this:
```yml
T-feat:
  - pr-title: "^feat"

T-fix:
 - pr-title: "^fix"
```
And get respective labels on PRs based on their titles.

**Justification:**
Allows automation for workflows that have PR naming conventions and we already have something similar with the branch names, So if there isn't an API restriction making it impossible it can be a useful addition.

**Are you willing to submit a PR?**
Yes

yyolk:
**Description:**
Adds an additional match configuration for any and all for matching PR titles.

If this PR is acceptable I can also add matching PR body content matching in the same way.

**Related issue:**

supersedes, closes #688
supersedes, closes #109 
refs #55 
closes #809 

**Check list:**
- [X] Mark if documentation changes are required. (Included)
- [X] Mark if tests were added or updated to cover the changes.

### DIFF
--- a/src/title.ts
+++ b/src/title.ts
@@ -1,0 +1,83 @@
+import * as core from '@actions/core';
+import * as github from '@actions/github';
+
+export interface TitleMatchConfig {
+  title?: string[];
+}
+
+export function toTitleMatchConfig(config: any): TitleMatchConfig {
+  if (!config['title']) {
+    return {};
+  }
+
+  const titleConfig = {
+    title: config['title']
+  };
+
+  if (typeof titleConfig.title === 'string') {
+    titleConfig.title = [titleConfig.title];
+  }
+
+  return titleConfig;
+}
+
+export function getTitle(): string | undefined {
+  const pullRequest = github.context.payload.pull_request;
+
+  if (!pullRequest) {
+    return undefined;
+  }
+
+  return pullRequest.title;
+}
+
+export function checkAnyTitle(regexps: string[]): boolean {
+  const title = getTitle();
+  if (!title) {
+    core.debug(`    cannot fetch title from the pull request`);
+    return false;
+  }
+
+  core.debug(`    checking "title" pattern against ${title}`);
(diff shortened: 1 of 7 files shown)


shadoath:
Beautiful! I would like to see this as well!

src/title.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  if (typeof titleConfig.title === 'string') {
+    titleConfig.title = [titleConfig.title];
+  }
+
+  return titleConfig;
+}
+
+export function getTitle(): string | undefined {
+  const pullRequest = github.context.payload.pull_request;
```
2bndy5, line 25: Can we instead use REST API (octokit) to get the title dynamically?

I think the `github.context` may be statically set upon workflow event trigger. This would lead to confusion about using an outdated PR title after it has changed without triggering new CI event (eg. re-running the job that uses this action after editing the PR title).

Yicong-Huang:
Would love to see this feature!

cameronmeissner:
@2bndy5 @yyolk anyway we can push this through? would greatly appreciate it

2bndy5:
I'm not an admin for this project. I just reviewed the patch for undesirable behavior, which still hasn't been addressed.

Revathyvenugopal162:
It seems like a useful feature. It would be great if it were available, especially when using conventional commits.

IchordeDionysos:
@loganrosen @chiranjib-swain it fixes the # 1 open issue in this repo.

We've been using a different action patching this together, though ...
... given other actions are less well maintained than this, it's quite annoying to push maintainers to update their actions, or to constantly fork all actions...

Do you see a chance to add support for labelling based on more factors, such as the title?

PR #867, opened 2025-05-01
docs: Update document to include issues permission

metarsit:
**Description:**
```
Error: HttpError: You do not have permission to create labels on this repository.: {"resource":"Repository","field":"label","code":"unauthorized"}
Error: You do not have permission to create labels on this repository.: {"resource":"Repository","field":"label","code":"unauthorized"}
```
Ran into this error, it seems that we have mission permission that's needed for the labeler to work.

Therefore, including the extra permission into the README.md so no one has to go through what I did. 😆 

**Related issue:**
https://github.com/orgs/community/discussions/156181

**Check list:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ The match object allows control over the matching options. You can specify the l
 
 The base match object is defined as:
 ```yml
-- changed-files: 
+- changed-files:
   - any-glob-to-any-file: ['list', 'of', 'globs']
   - any-glob-to-all-files: ['list', 'of', 'globs']
   - all-globs-to-any-file: ['list', 'of', 'globs']
@@ -125,7 +125,7 @@ Documentation:
 - changed-files:
   - any-glob-to-any-file: ['docs/*', 'guides/*']
 
-# Add 'Documentation' label to any change to .md files within the entire repository 
+# Add 'Documentation' label to any change to .md files within the entire repository
 Documentation:
 - changed-files:
   - any-glob-to-any-file: '**/*.md'
@@ -160,6 +160,7 @@ jobs:
     permissions:
       contents: read
       pull-requests: write
+      issues: write
     runs-on: ubuntu-latest
     steps:
     - uses: actions/labeler@v5
@@ -204,12 +205,13 @@ jobs:
     permissions:
       contents: read
       pull-requests: write
+      issues: write
     runs-on: ubuntu-latest
     steps:
-    
+
     # Label PRs 1, 2, and 3
     - uses: actions/labeler@v5
-      with:        
+      with:
         pr-number: |
           1
           2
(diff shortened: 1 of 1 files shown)


phun-ky:
Yeah, I ran into the same thing, but odd that you need issue permission to write to write labels to a PR. furthermore, I have two identical repositories, with seemingly identical workflows, but the latest repository fails, but the older one does not. All settings are checked, and all tokens are verified:

works: https://github.com/phun-ky/typeof/actions/runs/14731948751
does not work: https://github.com/phun-ky/eslint-config/actions/runs/14795721253

CommanderRedYT:
I have experienced the same thing, but even with the `issues` permission it does not work for me: https://github.com/deppenpfostierung/bicycle-led-software/actions/runs/14823033959/job/41612578804

McGeerDev:
This resolved the permission issues for me!

Jiralite:
For clarity, the reason why `issues: write` _may_ be needed is for when the label does not exist.

For example, if you have no "testing" label and you run:
```Zsh
curl -L \
  -X PUT \
  -H "Accept: application/vnd.github+json" \
  -H "Authorization: Bearer <token>" \
  -H "X-GitHub-Api-Version: 2022-11-28" \
  https://api.github.com/repos/<owner>/<repository>/issues/<number>/labels \
  -d '{"labels":["testing"]}'
```

Which is [this](https://docs.github.com/en/rest/issues/labels?apiVersion=2022-11-28#set-labels-for-an-issue) from the documentation and this from the source code:

https://github.com/actions/labeler/blob/b0a1180683c9f17424de4d71c044bea4c7b9bc7c/src/api/set-labels.ts#L1-L15

The label will be created and placed!

Those who are not experiencing problems are because they already have the labels in place. `issues: write` is only needed if labels are not already present.

Jiralite (commented):
For clarity, the reason why `issues: write` _may_ be needed is for when the label does not exist.

For example, if you have no "testing" label and you run:
```Zsh
curl -L \
  -X PUT \
  -H "Accept: application/vnd.github+json" \
  -H "Authorization: Bearer <token>" \
  -H "X-GitHub-Api-Version: 2022-11-28" \
  https://api.github.com/repos/<owner>/<repository>/issues/<number>/labels \
  -d '{"labels":["testing"]}'
```

Which is [this](https://docs.github.com/en/rest/issues/labels?apiVersion=2022-11-28#set-labels-for-an-issue) from the documentation and this from the source code:

https://github.com/actions/labeler/blob/b0a1180683c9f17424de4d71c044bea4c7b9bc7c/src/api/set-labels.ts#L1-L15

The label will be created and placed!

Those who are not experiencing problems are because they already have the labels in place. `issues: write` is only needed if labels are not already present, which is why a seemingly unrelated error occurs:

> HttpError: You do not have permission to create labels on this repository

That being said, can you please ensure this is mentioned somewhere in the README.md? I do not see this is present.

PR #923, opened 2026-02-10
Add 'changed-files-labels-limit' and 'max-files-changed' configs to allow capping number of labels added

bluca:
**Description:**
When a repository has many components, each with a changed-files label, a large refactor ends up with the labeler spamming the pull request with label changes. The end result is not very useful as it's not very readable, and due to how github automatically hides comments when label changes overflow the discussion tab, it means useful information is hidden and one has to manually click "Load more..." dozens of time every time the page is loaded.

Add a changed-files-limit top level config knob. If more than the configured limit of labels is set to be added, none are added. This only affects changed-files labels.

**Check list:**
- [X] Mark if documentation changes are required.
- [X] Mark if tests were added or updated to cover the changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,6 +160,7 @@ The labeler configuration file (`.github/labeler.yml`) supports the following to
 | Name                         | Description                                                                                                                                                   |
 |------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------|
 | `changed-files-labels-limit` | Maximum number of new labels to apply based on changed files (must be a non-negative integer). If exceeded, no changed-files labels are applied for that run. |
+| `max-files-changed`          | Maximum number of total changed files (must be a non-negative integer). If exceeded, all file-based labeling is skipped.                                      |
 
 ##### Limiting changed-files labels
 
@@ -172,6 +173,7 @@ When working with large PRs (e.g., tree-wide refactors) that touch many componen
 - If the number of new changed-files labels **equals** the limit, labels are still applied normally.
 - Labels based on branch conditions (`head-branch`, `base-branch`) are **not affected** by the limit.
 - **Any label definition that includes a `changed-files` rule is considered a changed-files label** and is subject to the limit, regardless of which condition caused the label to match. For example, a label with both `head-branch` and `changed-files` rules will be subject to the limit even if it matches via the branch rule.
+- If both `max-files-changed` and `changed-files-labels-limit` are configured at the same time, `max-files-changed` is evaluated first, and if it triggers, `changed-files-labels-limit` is not evaluated.
 
 ##### Example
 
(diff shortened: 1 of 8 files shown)


bluca:
Example of a PR where the labeling gets out of hands and spams the discussion page: https://github.com/systemd/systemd/pull/40624

bluca:
@chiranjib-swain anything else I can do here? Thanks

chiranjib-swain:
Hi @bluca ,Really nice work on this 👏 — the addition of `max-files-changed` is a great improvement and should help a lot in reducing label noise on large PRs.

**🔹 Suggestion: Improve parseNonNegativeInteger**

One small suggestion regarding `parseNonNegativeInteger` used here: https://github.com/bluca/labeler/blob/a07363b7e377762584a79fe87e6a744d5a385058/src/api/get-label-configs.ts#L33 

Currently the regex check runs directly on the raw string value. This means inputs like " 30 " would be rejected even though they represent a valid integer. It might be worth trimming the string before applying the regex so that common whitespace cases are handled more gracefully.

Since the regex already guarantees that the string contains only digits, we could also use `Number(trimmed)` instead of `parseInt`. That avoids relying on parseInt’s permissive behavior while keeping the logic clear.

Finally, the non-negativity check (`value < 0`) could potentially be included directly in the numeric branch condition to simplify the validation flow slightly.

Something along these lines could work:

```
const trimmed = value.trim()

if (!/^\d+$/.test(trimmed)) {
  throw new Error(
    `Invalid value for '${optionName}': must be a non-negative integer (got '${value}')`
  )
}

parsed = Number(trimmed)
```

This is a fairly small refinement, but it could make the validation a bit more robust while keeping the current behavior intact.

**🔹 Suggestion: Clarify Threshold Semantics**

Additionally, one small suggestion regarding threshold semantics:

If both limits are introduced/configured (e.g., `changed-files-max-files` and `changed-files-labels-limit`), it might help to clearly define their interaction. For example:

> When both are configured, `changed-files-max-files` is evaluated first. If it triggers, `changed-files-labels-limit` is not evaluated.

Documenting this precedence explicitly could help avoid confusion and make the behavior more predictable.

**🔹 Suggestion: Reserved Keys Clarification**

Using keys like `max-files-changed` as labels in `.github/labeler.yml` causes failures because they are treated as reserved config options and parsed as integers.

Currently, this results in an error like:

> Invalid value for 'max-files-changed': expected a non-negative integer

We should either document these as reserved keys or improve the error message, e.g.:

> 'max-files-changed' is a reserved top-level option and cannot be used as a label. Please rename it.

**🔹 Suggestion: PR Title Improvement**

Your current title:
Add `'changed-files-labels-limit'` and `'changed-files-labels-limit'` configs to allow capping number of labels added

has a typo (the same option is repeated) and it doesn’t mention `max-files-changed`.

bluca:
@chiranjib-swain thank you, updated as requested, PTAL

chiranjib-swain (approved):

Issue #763, opened 2024-03-17
Labeler should never remove labels it has not configured


silverwind:
**Description:**
The label addition is not atomic and there is a small time window after opening a PR where the action will unexpectedly remove user-added labels. Example from [here](https://github.com/go-gitea/gitea/pull/29856):

<img width="893" alt="image" src="https://github.com/actions/labeler/assets/115237/d120c27d-d993-4a11-94ec-6f509aeda15b">

I guess the API flow is like this:

- Labeler does a `GET labels` request
- User changes labels (which invalidates the labels seen in the GET)
- Labeler does a `PUT labels` request with previously seen labels, reverting the user's changes

**Action version:**
`v5`

**Platform:**
- [x] All

**Runner type:**
- [x] All

**Repro steps:**  
1. Have a simple labeler action with [this config](https://github.com/go-gitea/gitea/blob/ed02d1fab85c9b8206c0af84dcfc3792e61609cf/.github/workflows/pull-labeler.yml#L11-L20)
2. Open a PR
3. Immediatly set some labels on the PR
4. If the timing is right, the labeler will unexpectedly remove the user-added labels

**Expected behavior:**
The action should atomically update the labels or at least never removing labels it does not have in its config.

**Actual behavior:**
The action will remove user-set labels when set within a certain time window.

HarithaVattikuti:
Hello @silverwind 
Thank you for creating this issue. We will investigate it and get back to you as soon as we have some feedback.

silverwind:
I reworded the PR because I think "atomic" is not the right term. Essentially it should be made to never remove labels that it are not in the configuration. It will need to use a 1-shot API to achieve this, as 2-shot will always have the race condition.

nzakas:
We are running into the same issue, where the labeler is removing a label that a different bot added. 
https://github.com/eslint/eslint/issues/18183
<img width="874" alt="311469955-abb37dcb-fa01-4be9-84e9-b0bc5aac8583" src="https://github.com/actions/labeler/assets/38546/7c7c21d7-4c67-4421-80b0-6784a2bd311a">

jorgepiloto:
Until this gets fixed, my approach has been registering the labels with a glob pattern that will never match a file (unless the only modified file is named 'THIS-NEVER-MATCHES-A-FILE'):

```yml

'your-label':
  - all:
      - changed-files:
          - all-globs-to-all-files: ['THIS-NEVER-MATCHES-A-FILE']
```

[...]

Issue #561, opened 2023-05-10
HttpError: Server Error when using labeler with over 50 labels


jkinnula-elisa:
**Description:**
Do a labeler configuration in a mono-repo with over 50 components. The end result will be `HttpError: Server Error` due labeler posting labels update twice, on which first time succeeds - and second time fails.

GitHub has limit of max 100 labels on a Pull Request. On this case we are not talking about a mono-repo which has over 100 components (which would result 100 or over 100 labels), but a case that labeler does update the labels - and in the next update does post the labels again. GitHub does sanity check for addLabels - not to be able to exceed possible amount of new labels (e.g. if there's 51 labels already in PR, possible amount of new labels is 49). On this case the labels are actually the same, since labeler posted already the same labels before.

For short, this bug is about that error situation will happen in case that PR has more than 50 labels.

**Action version:**
v4.0.3

**Platform:**
- [X] Ubuntu
- [X] macOS
- [X] Windows

**Runner type:**
- [X] Hosted
- [X] Self-hosted

**Repro steps:**  
- Create labeler configuration with 51 labels, merge the configuration into main/master
- Create a test PR which touches files enough to make 51 labels

**Expected behavior:**
No errors, and 51 labels are added.

**Actual behavior:**
Labels are added, but labeler workflow run failed due to error `HttpError: Server Error`.

MaksimZhukov:
Hello @jkinnula-elisa ! 
Thank you for reporting this! We will investigate the issue

PR #957, opened 2026-07-20
Preserve externally managed labels during synchronization

loganrosen:
**Description:**
`sync-labels` currently narrows its race window with a final label read, then replaces the complete label set. A concurrent label added after that read can still be lost.

This restores the selective label-ownership isolation used by v4 before #497: matching configured labels are added in one REST batch, while stale configured labels are removed in one selective GraphQL mutation. It retains v6 matching, synchronization, and label-limit semantics without ever rewriting unconfigured labels. Addition retries are disabled; if a server error follows a committed write, the action checks the current labels, following pagination as needed, before deciding whether to surface the original failure.

**Related issue:**
Addresses #763.

**Check list:**


v-chiranjib-swain:
<p>Thanks for working on this and for investigating the non-atomicity issue.</p>

<p>I spent some time testing this approach with repositories configured with a large number of labels, and I wanted to share a few observations that may be worth considering.</p>

<p>In my testing, <code>addLabels</code> (POST) became unreliable as the number of labels increased. For example, I observed:</p>

<ul>
  <li><strong>502 Bad Gateway</strong> when attempting to add <strong>75 labels</strong> (and in some <strong>100-label</strong> scenarios).</li>
  <li>In the 502 case, GitHub had already committed the complete label update server-side, but still returned a 502 to the client. From the action's perspective, this is indistinguishable from a genuine write failure, so the workflow reports a failure even though the desired state has already been reached.</li>
</ul>

<hr>

<p>Because of that, I'm not convinced that <code>retries: 0</code> fully addresses the problem. While it avoids retrying a non-idempotent request, it also means users receive a failed workflow for an operation that may have already completed successfully.</p>

<p>Since the labels have already been applied, <code>newLabels</code> is empty on the next workflow run, so the action succeeds without performing any work. However, this only happens if another workflow run is triggered (for example, by pushing a new commit or manually re-running the workflow). Otherwise, the PR continues to display the failed check even though the desired state has already been reached.</p>

<p>In that sense, there is nothing to "<strong>heal</strong>"—the initial failure is a <strong>false negative</strong>.</p>

<hr>

<p>Given that
<a href="https://github.com/actions/labeler/pull/497">#497</a>
introduced <code>setLabels</code> primarily to enforce the 100-label limit—and that an effective side-effect was avoiding these POST failure modes—reverting to <code>addLabels</code> reintroduces the same reliability problems documented in
<a href="https://github.com/actions/labeler/issues/561">#561</a>.
I think it's worth considering whether this trade-off is appropriate for repositories with a large number of configured labels.</p>

<hr>

<p>I've attached a summary of the workflow failure from my testing, highlighting the failed GitHub PR check and the corresponding 502 Server Error logs.</p>

<p>I hope these findings are helpful in evaluating the trade-offs of this approach.</p>
<p>
  Workflow failure summary
</p>

loganrosen:
Thanks for reproducing this—you’re right that `retries: 0` alone still leaves a false-negative failure when the POST commits before returning 502. I’ve added read-after-error reconciliation: only after a 5xx, the action fetches the current labels once with retries disabled and treats the addition as successful only if every requested label is present. If any label is missing or verification fails, it preserves the original POST error. The normal path remains one request, with no blind POST retry. Regression coverage now includes the committed-502, incomplete-write, verification-failure, and non-5xx cases.

v-chiranjib-swain:
<p>Thanks for this PR — the core goal is well-implemented and verified. I've done end-to-end testing at scale (100-label config, simultaneous comparison runs against main).</p>

<h3>Two things worth addressing before merge</h3>

<h4>1. README documentation — transient zero-label state</h4>

<p>When <code>sync-labels: true</code> triggers a label swap (stale labels removed, new labels added), there is a ~10s window between <code>removeLabels</code> completing and <code>addLabels</code> completing where the PR has zero config labels. This is an inherent consequence of the two-call design and not a bug — but users will encounter it and have no way to understand it from the current docs.</p>

<p><strong>Suggested README addition under a <code>sync-labels</code> note:</strong></p>

<blockquote>
  <p>When <code>sync-labels: true</code> triggers a label swap, there is a brief window (~10s) where the PR has no config-matched labels while the operation is in progress. Any webhook or CI system reading labels during this window will see zero labels.</p>

  <p>Additionally, for repos using 100 config labels (at the GitHub hard cap): if an external label is added during this window, <code>addLabels</code> will hit the cap. GitHub may respond with either:</p>

  <p><strong>422 Validation Failed</strong> — the entire POST is rejected pre-write; all 100 config labels are lost and only the injected label remains on the PR. A re-run does not fully resolve this — with the injected label still present, the labeler's 100-label cap will silently exclude one config label as excess (logged as <code>Warning: Maximum of 100 labels allowed</code>). Full resolution requires removing the injected label manually first, then re-running.</p>

  <p><strong>502 Server Error</strong> — the write commits server-side (all 100 config labels stored), but the reconciliation's single-page GET misses any label overflowing to REST page 2; labeler reports failure despite the write succeeding. A re-run fully resolves this state.</p>

  <p>Risk threshold: <code>config_labels + concurrent_injections &gt; 100</code> — only realistic at the 100-label cap itself.</p>
</blockquote>

<h4>2. Performance regression on <code>sync-labels</code> swap path</h4>

<p>On the remove+add path, PR #957 is ~2x slower than main (35s vs 19s in simultaneous comparison runs, cross-verified at 37s in a second independent run).</p>

<ul>
  <li>
    <strong>Add labels only (fresh):</strong>
    <ul>
      <li>pr-957: ~18s</li>
      <li>main: ~20s</li>
    </ul>
  </li>

  <li>
    <strong>Remove + Add (swap):</strong>
    <ul>
      <li>pr-957: ~35s</li>
      <li>main: ~19s</li>
    </ul>
  </li>
</ul>

<p>Root cause: two sequential API calls (<code>removeLabelsFromLabelable</code> GraphQL → <code>addLabels</code> POST) vs main's single atomic <code>setLabels</code> PUT. This is an inherent trade-off of the approach — worth documenting so users with time-sensitive workflows are aware.</p>

loganrosen:
Thanks for the detailed scale testing. I addressed these concerns in `eb228e3`:

- The 5xx reconciliation now follows REST pagination while keeping retries disabled on every GET, so a requested label pushed to page 2 by a concurrent cap race is recognized. The regression covers 100 labels on page 1 and the final requested label on page 2.
- The README now documents the sequential remove/add behavior: configured labels may be temporarily absent during a swap, swaps can take longer, and an external label added at the 100-label cap can prevent all configured additions because labeler will not remove that external label to make room.

I kept the performance wording qualitative because the exact duration and multiplier vary by environment.

chiranjib-swain (approved):

priya-kinthali (approved):

HarithaVattikuti (approved):

Issue #434, opened 2022-09-06
Please support minimatch options 


nefrob:
Ex. I want to match dotfiles by adding `dot: true` option to my glob pattern to minimatch.

kachkaev:
Possible duplicate of https://github.com/actions/labeler/issues/135

nefrob:
@kachkaev yes it appears so. Closing, but it would be great to get the above PR approved and merged. Thanks.

Issue #135, opened 2021-04-06
Action does not match folders starting with dot (.)


ibratoev:
Actual:
* if I have a change in a folder `src/.deploy` and matcher for `src/**/*` it would not be matched.

Expected:
* folders starting with dot should be matched as well.

It seems the issue is related to how Minimatch works. See https://github.com/isaacs/minimatch/issues/30 . Consider passing `{dot: true}` by default.

alfieyfc:
I'm just giving it a wild guess, would it be just to add that parameter to these two lines?

```ts
const matchers = globs.map(g => new Minimatch(g, {dot: true})); 
```

https://github.com/actions/labeler/blob/69990a6043bd2b78ec675ec775d8e1cb62204ff0/src/main.ts#L187
https://github.com/actions/labeler/blob/69990a6043bd2b78ec675ec775d8e1cb62204ff0/src/main.ts#L202

mcieplucha:
actually I'm having issues event with root level configuration like:
`- ./*.js`
which seems to be not working. anyone did managed to make it work?

lPaths:
been 5 months already, any update on this?

boliveira:
How is the status of this?

beefchimi:
Would also love to see a fix for this 🙏

logicbomb421:
My company uses this action and is also encountering this issue. Seems like this is a pretty simple update, would it be possible to get this out soon?

kachkaev:
Today we were [configuring labels](https://github.com/blockprotocol/blockprotocol/pull/191) in [blockprotocol/blockprotocol](https://github.com/blockprotocol/blockprotocol) and I was like ‘on no not this dot problem again’ 😅

So here is the fix: https://github.com/actions/labeler/pull/316. It is non-breaking, just allows us to set the `dot` parameter:

```diff
 name: "Pull Request Labeler"
 on:
 - pull_request_target

 jobs:
   triage:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/labeler@v3
       with:
         repo-token: "${{ secrets.GITHUB_TOKEN }}"
+        dot: true
```

I hope it will be available for everyone soon! 🤞

kachkaev:
The solution to this confusing issue has been up for nearly 10 months: https://github.com/actions/labeler/pull/316 😔 Any chance we can get it merged before NY2023? 🤞

MaksimZhukov:
Hello everyone! 
[...]

PR #978, opened 2026-09-15
feat: add `nocase` input for case-insensitive glob matching

### LINKED ISSUE #912: Case sensitive matching?
Could you document if case sensitive matching is done?

And if not, how it matching could be done in a case insensitive matter?

My use case is that I want to match files containing `cmake` or `CMake`.

freezysko:
**Description:**

Adds a `nocase` action input that is passed through to minimatch, in exactly the same way the existing `dot` input already is. It controls whether `changed-files` glob patterns match paths case-insensitively.

```yaml
- uses: actions/labeler@v6
  with:
    nocase: true
```

```yaml
# .github/labeler.yml
my-component:
  - changed-files:
      - any-glob-to-any-file: ['src/mycomponent/**']
```

With `nocase: true` the pattern above also matches `src/MyComponent/Button.tsx`.

The input defaults to `false`, so existing configurations keep their current case-sensitive matching behaviour and nothing changes for anyone who does not opt in.

**Scope / non-goals:**

`nocase` only affects `changed-files` glob patterns, because those are the only patterns evaluated by minimatch. `head-branch` and `base-branch` patterns are compiled with `new RegExp()` in `src/branch.ts` and are therefore unaffected. This is stated explicitly in the README so it does not read as a bug later. Case-insensitive branch matching is a separate change — see #836, which touches `src/branch.ts` only and does not overlap with this PR.

**Related issue:**

Fixes #912 — the request there is to match e.g. both `cmake` and `CMake`, and a comment on the issue points directly at [minimatch's `nocase` option](https://github.com/isaacs/minimatch?tab=readme-ov-file#nocase) while asking how to enable it. This PR exposes exactly that option.

This also follows the precedent set by #135 / #434, where the `dot` minimatch option was exposed as an action input.

**Implementation:**

- `action.yml` — new `nocase` input, `default: false`, `required: false`
- `src/get-inputs/get-inputs.ts` — reads the input
- `src/labeler.ts` — threads the value through `checkMatchConfigs` → `checkMatch` → `checkAny` / `checkAll` → `checkAnyChangedFiles` / `checkAllChangedFiles`
- `src/changedFiles.ts` — threads it to all four `new Minimatch(g, {dot, nocase})` call sites
- `README.md` — documents the input and its scope
- `dist/index.js` — rebuilt

The value is passed as a positional boolean next to `dot` to keep the diff minimal and to mirror the existing convention.

**Tests:**

Every test asserts **both** directions on the same pattern/path pair, where letter case is the only difference — a test that only asserted the `nocase: true` direction would still pass if the flag were hardcoded.

`__tests__/changedFiles.test.ts` runs a scenario table against all four `checkIf*` matching functions:

| Scenario | Changed files | Glob patterns |
| --- | --- | --- |
| CamelCase directory name | `src/MyComponent/Button.tsx`, `src/MyComponent/Icon.tsx` | `src/mycomponent/**`, `SRC/mycomponent/*.tsx` |
| CamelCase file name | `src/MyComponent.tsx`, `src/MyHelper.tsx` | `src/my*.tsx`, `src/MY*.TSX` |
| Uppercased file extension | `src/foo.md`, `src/bar.md` | `**/*.MD`, `SRC/*.md` |
| Lowercase paths, CamelCase patterns | `docs/readme.md`, `docs/contributing.md` | `Docs/*.md`, `docs/*.Md` |

It additionally covers every `changed-files` config key (`any-glob-to-any-file`, `any-glob-to-all-files`, `all-globs-to-any-file`, `all-globs-to-all-files`), since `checkAnyChangedFiles` short-circuits on the first matching key and would otherwise leave branches untested.

`__tests__/labeler.test.ts` covers `checkMatchConfigs`, and `__tests__/main.test.ts` covers the full path through `run()` with a new `camel_case_paths.yml` fixture, verifying the input is propagated from the action input all the way to minimatch.

All 176 tests pass, and `format-check`, `lint`, `tsc` and `build` are clean.

**Check list:**

---

🤖 Generated with [Claude Code](https://claude.com/claude-code)


oga35767-eng (approved):

oga35767-eng:
👍

